PR #2, opened 2017-09-11
make suite.out user-friendly

sharkcz:
Put some separators to the test cases outputs so suite.out
is more readable.


p-steuer:
Merged cf7bbe84f6a82226553ac6d751754d5242d94bd1

PR #5, opened 2017-11-27
Ecc additions

p-steuer:


libica.map
```diff
@@ -117,5 +117,7 @@ LIBICA_3.2.1 {
 	ica_ecdh_derive_secret;
```
p-steuer: we ll do a 3.3.0 (instead of 3.2.1), because of the new apis

src/ica_api.c
```diff
@@ -1242,6 +1242,28 @@ int ica_ecdsa_verify(ica_adapter_handle_t adapter_handle,
 	return rc;
 }
 
+int ica_ec_key_get_public_key(ICA_EC_KEY *key, unsigned char *q, unsigned int *q_len)
```
p-steuer: can key parameter be qualified by const? it seems to be read only in this context.

src/ica_api.c
```diff
@@ -1242,6 +1242,28 @@ int ica_ecdsa_verify(ica_adapter_handle_t adapter_handle,
...
+    if (!key || !(key->X) || privlen_from_nid(key->nid) < 0)
+	return EINVAL;
+
+    memcpy(q, key->X, 2*privlen_from_nid(key->nid));
+    *q_len = 2*privlen_from_nid(key->nid);
+
+    return 0;
+}
+
+int ica_ec_key_get_private_key(ICA_EC_KEY *key, unsigned char *d, unsigned int *d_len)
```
p-steuer: can key parameter be qualified by const? it seems to be read only in this context.

src/s390_ecc.c
```diff
@@ -468,7 +468,8 @@ unsigned int make_ecdsa_verify_parmblock(char *pb,
  * makes an ECDSA key structure at given struct and returns its length.
  */
 unsigned int make_ecdsa_private_key_token(unsigned char *kb,
-		const ICA_EC_KEY *privkey, uint8_t curve_type)
+		const ICA_EC_KEY *privkey, unsigned char *X, unsigned char *Y,
```
p-steuer: can X and Y parameters be ualified be const? they seem to be read only. also applies to make_ecdsa_sign_request

src/s390_ecc.c
```diff
@@ -468,7 +468,8 @@ unsigned int make_ecdsa_verify_parmblock(char *pb,
...
+		uint8_t curve_type)
```
p-steuer: is there a reason to restrict the curve type to 8bit. why not just use an int?

src/s390_ecc.c
```diff
@@ -598,12 +602,96 @@ ECDSA_SIGN_REPLY* make_ecdsa_sign_request(const ICA_EC_KEY *privkey,
...
+	bn_x = BN_new();
+	bn_y = BN_new();
+	if (!EC_POINT_get_affine_coordinates_GFp(group, pub_key, bn_x, bn_y, NULL)) {
+		rc = EFAULT;
+		goto end;
+	}
+
+	/* Format (X) as char array, with leading zeros if necessary */
+	x_str = BN_bn2hex(bn_x);
+	n = privlen - strlen(x_str) / 2;
```
p-steuer: is this guaranteed to be divisible by two? if not: does this still work ?

src/s390_ecc.c
```diff
@@ -598,12 +602,96 @@ ECDSA_SIGN_REPLY* make_ecdsa_sign_request(const ICA_EC_KEY *privkey,
...
+	for (i = 0; i < n; i++)
+		X[i] = 0x00;
+	BN_bn2bin(bn_x, &(X[n]));
+
+	/* Format (Y) as char array, with leading zeros if necessary */
+	y_str = BN_bn2hex(bn_y);
+	n = privlen - strlen(y_str) / 2;
```
p-steuer: same as above.

src/s390_ecc.c
```diff
@@ -759,19 +850,18 @@ unsigned int ecdsa_verify_sw(const ICA_EC_KEY *pubkey,
...
     }
 
-    if (!EC_KEY_check_key(a)) {
-        goto err;
-    }
-
     /* create ECDSA_SIG instance */
     sig = ECDSA_SIG_new();
     r = BN_bin2bn(signature, privlen, r);
     s = BN_bin2bn(signature+privlen, privlen, s);
```
p-steuer: BN_bin2bn allocates BIGNUM structures for both s and r. The memory should be freed before return.

src/ica_api.c
jschmidb: Yes! Will do ...

src/ica_api.c
jschmidb: Yep, same as for get_public_key

src/s390_ecc.c
jschmidb: Yes, will do.

src/s390_ecc.c
jschmidb: Yes, there is a reason .... the curve_type field in the ECC_PRIVATE_KEY_SECTION is a 8-bit value: 
kp1->privsec.curve_type = curve_type;
If curve_type has already the right size, we don't need a cast. In fact I'm "enforcing" this cast by passing the curve_type parm to make_ecdsa_private_key_token(). In the calling routine make_ecdsa_sign_request() curve_type is in fact a int. But I could change the curve_type_from_nid() routine to return a uint8_t instead of an int, so that we have always the right type.

p-steuer: okay. if the curve_type is a 8-bit exact-width int, id suggest to use it uint8_t consistently.

src/s390_ecc.c
jschmidb: Yes .... for example, when using the p521 curve, x_str sometimes has only 65 bytes (Could be even less, depending on the actual value of bn_x). Dividing by 2 results in a rounded int of 32 bytes. However .... I think this division is not necessary at all. What I want to do here, is pre-pending 0x00 bytes so that the resulting char array for X has exactly 66 bytes in this example. The code works, but does too much. It would work also without dividing by 2:  

x_str = BN_bn2hex(bn_x);    // <- e.g. 65 bytes
n = privlen - strlen(x_str);    // <- n = 66 - 65 = 1
for (i = 0; i < n; i++)           
	X[i] = 0x00;                  // <- X[0] = 0x00
BN_bn2bin(bn_x, &(X[n]));   // <- place the bn_x value at X[1] 
  
What do you think?

src/s390_ecc.c
jschmidb: Same as above.

src/s390_ecc.c
jschmidb: I'm calling ECDSA_SIG_free before return and checked again what ECDSA_SIG_free does .... Here is the code from ec_asn1.c: 

void ECDSA_SIG_free(ECDSA_SIG *sig)
{
    if (sig == NULL)
        return;
    BN_clear_free(sig->r);
    BN_clear_free(sig->s);
    OPENSSL_free(sig);
}

So ECDSA_SIG_free also frees r and s.

p-steuer: ok

src/s390_ecc.c
p-steuer: No, the division by 2 was needed because 1 byte is encoded as 2 hex digits. Question is why do you encode the binary value to a hex *string*, measure the length of that string and divide it by two to get the number of bytes instead of using BN_num_bytes(bn_x)? i didnt get it.

src/ica_api.c
hfreude: Can we improve the interface so that it's possible to call the function(s) first without
buffer to query for the buffer size:
int ica_ec_..._key(ICA_EC_KEY *key, unsigned char *q, unsigned int *q_len)
{
    ...
    if (q)
        memcpy(q, ...);
    *q_len = 2*privlen_from_nid(...);
    ...
}

src/s390_ecc.c
hfreude: I generally I vote for int or unsigned int here and just doing the cast the moment you fill the CPRB struct.
This way the API has room for future extensions, it is narrowed down exactly where the one and only
hardware is used. But new hardware (e.g. ECC on CPACF) may not have this restrictions and may support
future ECC curves...

src/s390_ecc.c
```diff
@@ -559,12 +560,15 @@ unsigned int make_ecdsa_public_key_token(ECDSA_PUBLIC_KEY_BLOCK *kb,
...
 /**
- * creates an ECDSA xcrb request message for zcrypt.
+ * creates an ECDSA sign request message for zcrypt. The given private key does usually
+ * not contain a public key (X,Y), but the card requires (X,Y) to be present. The
+ * calling function makes sure that (X,Y) are correctly set.
  *
  * returns a pointer to the control block where the card
  * provides its reply.
  */
 ECDSA_SIGN_REPLY* make_ecdsa_sign_request(const ICA_EC_KEY *privkey,
```
hfreude: static ?

src/include/s390_ecc.h
```diff
@@ -245,6 +245,7 @@ typedef struct {
 } __attribute__((packed)) ECDSA_VERIFY_REPLY;
 
 ECDSA_SIGN_REPLY* make_ecdsa_sign_request(const ICA_EC_KEY *privkey,
```
hfreude: static ?

src/s390_ecc.c
```diff
@@ -598,12 +602,96 @@ ECDSA_SIGN_REPLY* make_ecdsa_sign_request(const ICA_EC_KEY *privkey,
...
+	offset += make_ecdsa_private_key_token(cbrbmem+offset, privkey, X, Y, curve_type);
 	finalize_xcrb(xcrb, preqcblk, prepcblk);
 
     return (ECDSA_SIGN_REPLY*)prepcblk;
 }
 
+/**
+ * calculate the public (X,Y) values for the given private key, if necessary.
+ */
+unsigned int provide_pubkey(const ICA_EC_KEY *privkey, unsigned char *X, unsigned char *Y)
```
hfreude: static ?

src/s390_ecc.c
```diff
@@ -598,12 +602,96 @@ ECDSA_SIGN_REPLY* make_ecdsa_sign_request(const ICA_EC_KEY *privkey,
...
+{
+	EC_KEY *eckey = NULL;
+	EC_POINT *pub_key = NULL;
+	const EC_GROUP *group = NULL;
+	BIGNUM *bn_d = NULL, *bn_x = NULL, *bn_y = NULL;
+	char* x_str = NULL;
+	char* y_str = NULL;
+	int privlen = -1;
```
hfreude: is used as lvalue in line 629, so no initialization value needed.

src/s390_ecc.c
```diff
@@ -598,12 +602,96 @@ ECDSA_SIGN_REPLY* make_ecdsa_sign_request(const ICA_EC_KEY *privkey,
...
 
+/**
+ * calculate the public (X,Y) values for the given private key, if necessary.
+ */
+unsigned int provide_pubkey(const ICA_EC_KEY *privkey, unsigned char *X, unsigned char *Y)
+{
+	EC_KEY *eckey = NULL;
+	EC_POINT *pub_key = NULL;
+	const EC_GROUP *group = NULL;
+	BIGNUM *bn_d = NULL, *bn_x = NULL, *bn_y = NULL;
```
hfreude: all these 3 are used as lvalues and so there is not initialization value needed.

src/s390_ecc.c
```diff
@@ -598,12 +602,96 @@ ECDSA_SIGN_REPLY* make_ecdsa_sign_request(const ICA_EC_KEY *privkey,
...
+	char* x_str = NULL;
+	char* y_str = NULL;
+	int privlen = -1;
+	unsigned int i, n, rc;
+
```
hfreude: I can't see the point why here all the variables get an initialization value. I checked the BIGNUM, privlen and eckey variable and their first use is as lvalue so no init needed.

src/s390_ecc.c
```diff
@@ -616,11 +704,19 @@ unsigned int ecdsa_sign_hw(ica_adapter_handle_t adapter_handle,
...
 	ECDSA_SIGN_REPLY* reply_p;
 	unsigned int privlen = privlen_from_nid(privkey->nid);
+	unsigned char X[MAX_ECC_PRIV_SIZE];
+	unsigned char Y[MAX_ECC_PRIV_SIZE];
 
 	if (adapter_handle == DRIVER_NOT_LOADED)
 		return EFAULT;
 
-	reply_p = make_ecdsa_sign_request(privkey, hash, hash_length, &xcrb);
+	rc = provide_pubkey(privkey, (unsigned char*)&X, (unsigned char*)&Y);
```
hfreude: are you sure this is (unsigned char*) &X ?
Don't you want to write
  rc = provide_pubkey(privkey, X, Y);
instead ?

src/s390_ecc.c
```diff
@@ -616,11 +704,19 @@ unsigned int ecdsa_sign_hw(ica_adapter_handle_t adapter_handle,
...
+	if (rc != 0)
+		return EFAULT;
+
+	reply_p = make_ecdsa_sign_request((const ICA_EC_KEY*)privkey,
+			(unsigned char*)&X, (unsigned char*)&Y,
```
hfreude: same here, use simple X and Y instead of (unsigned char *)&X

src/s390_ecc.c
```diff
@@ -646,7 +742,6 @@ unsigned int ecdsa_sign_sw(const ICA_EC_KEY *privkey,
 {
 	int rc = 1;
     EC_KEY *a = NULL;
-    BIGNUM* xa=NULL; BIGNUM* ya=NULL;
     BIGNUM* r=NULL; BIGNUM* s=NULL;
```
hfreude: Later in this function there is a line
  ECDSA_SIG_get0(sig, (const BIGNUM**)&r, (const BIGNUM **)&s);
can you please change this to
  ECDSA_SIG_get0(sig, &r, &s);

p-steuer:
Can you take a look again at the frees? Some functions like (BN_clear_free) can be passed a NULL pointer and they ll do nothing while others (like OPENSSL_free) will actually do a free on the NULL pointer.

src/s390_ecc.c
jschmidb: Agreed. BN_num_bytes() will do it.

src/ica_api.c
jschmidb: The current API is designed to return the private key's D-length when creating the key: 
ICA_EC_KEY* ica_ec_key_new(unsigned int nid, unsigned int *privlen);
So the application has this length when the key is created. 
We could improve the ica_ec_key_get_public_key() and ica_ec_key_get_private_key() functions to also provide this length, but it is not really necessary.

src/s390_ecc.c
jschmidb: Agreed. I will remove the initializations (and also go over the complete code again to find similar places).

p-steuer:
Rework is PR #6 . Closed.

PR #6, opened 2017-12-18
Libica ECC enhancements for Opencryptoki / icatoken.

jschmidb:
Several changes that are necessary for using Libica ECC in Opencryptoki / icatoken.


p-steuer:
Moved PR to internal repo. Close.

PR #9, opened 2018-02-09
New API function to set libica fallback mode.

hfreude:
Introduce a new function to set libica's fallback mode.
Slight rework for all the hw functions which do not
have fallbacks to return ENODEV. The old code used
sometimes EPERM, 1 and on some places already ENODEV.

Signed-off-by: Harald Freudenberger <freude@linux.vnet.ibm.com>

### DIFF
--- a/include/ica_api.h
+++ b/include/ica_api.h
@@ -426,6 +426,14 @@ typedef struct ica_drbg ica_drbg_t;
 #define ICA_FALLBACKS_DISABLED 0
 
 /**
+ * Environment variable for defining the default Libica fallback mode.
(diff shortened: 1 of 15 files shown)


p-steuer:
looks good.

- should ibmca use the new api to disable fallbacks? i think yes, but we have to be careful with libica - ibmca version dependencies.
- dont forget to add the new interface to the libica book
- should we additionally introduce a corresponding environment variable? It could be used to set system-wide defaults (that would be overwritten by application's settings).

jschmidb:
Let's remember that we already have the ICAPATH environment variable. Setting ICAPATH=1 causes libica to fail if the hardware path fails.

PR #10, opened 2018-02-14
Libica introduce fallback mode, code simplifications and cleanup 

hfreude:
- New API function to set libica fallback mode.
- Add libica fallback environment variable.
- Whitespace fix, cleanup and beautify switch code in s390_crypto.


p-steuer:
Moved to internal repo. Close.

PR #21, opened 2018-06-12
Add non-executable gnu stack markings in the assembly.

p-steuer:


sharkcz:
Patrick, what would you say to
````
diff --git a/src/Makefile.am b/src/Makefile.am
index c7cecf6..96fd2e6 100644
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -8,7 +8,7 @@ lib_LTLIBRARIES = libica.la
 
 libica_la_CFLAGS = ${AM_CFLAGS} -I${srcdir}/include -I${srcdir}/../include \
                   -fvisibility=hidden
-libica_la_CCASFLAGS = ${AM_CFLAGS}
+libica_la_CCASFLAGS = ${AM_CFLAGS} -Wa,--noexecstack
 libica_la_LIBADD = @LIBS@ -lrt -lcrypto
 libica_la_LDFLAGS = -Wl,--version-script=${srcdir}/../libica.map \
                    -version-number ${VERSION}

````
IMHO it would be better to have the option closer to its usage, rather than make it global in CFLAGS.

p-steuer:
im in favour of disabling all potential security risks globally per default and explicitely "white-list" the places where you cannot get around it.

So do you see a reason to not disable execstack globally ie. also for bins, tescases .. ?

sharkcz:
AFAIK the problem with executable stack is always in hand-written/generated assembler code and it's possible to fix it there too (set proper directives or such inside the source). So using --noexecstack globally (or locally) masks the real problem. Google found https://wiki.gentoo.org/wiki/Hardened/GNU_stack_quickstart#How_to_fix_the_stack_.28in_practice.29 for example.

p-steuer:
The article mentioned --noexecstack not being portable to all gnu systems, so its a convincing argument to fix it in the assembly.

pushed

PR #28, opened 2019-06-17
Devel

jschmidb:
Misc. housekeeping and bug fixes: set dynamic ICA flag according to available online cards, don't call ioctl for ECC functions if ECC is not available via cards, enhance icainfo output to three columns, distinguishing between static and dynamic hardware.

### DIFF
--- a/include/ica_api.h
+++ b/include/ica_api.h
@@ -68,10 +68,28 @@ typedef ica_adapter_handle_t ICA_ADAPTER_HANDLE;
 #define ICA_CBCCS_VARIANT2	2
 #define ICA_CBCCS_VARIANT3	3
 
+/**
+ * ICA flags
+ */
 #define ICA_FLAG_SHW 4 /* static hardware support (symmetric ops - CPACF) */
 #define ICA_FLAG_DHW 2 /* dynamic hardware support (asymmetric ops - CEX) */
 #define ICA_FLAG_SW  1 /* software implementation (fallback / backup) */
 
+/**
+ * ICA properties: key lengths
+ */
+#define ICA_PROPERTY_AES_128		0x00000001
+#define ICA_PROPERTY_AES_192		0x00000002
+#define ICA_PROPERTY_AES_256		0x00000004
+
+#define ICA_PROPERTY_RSA_ALL		0x0000000F /* All RSA key lengths */
+#define ICA_PROPERTY_EC_BP			0x00000001 /* Brainpool curves */
+#define ICA_PROPERTY_EC_NIST		0x00000002 /* NIST curves */
+#define ICA_PROPERTY_EC_ED			0x00000004 /* Edwards curves */
+
+/**
+ * Algorithms
+ */
 #define SHA1            1
 #define SHA224          2
 #define SHA256          3
--- a/src/ica_api.c
+++ b/src/ica_api.c
@@ -1017,7 +1017,11 @@ unsigned int ica_rsa_mod_expo(ica_adapter_handle_t adapter_handle,
 		rc = ica_fallbacks_enabled ?
 			rsa_mod_expo_sw(&rb) : ENODEV;
 	else {
-		rc = ioctl(adapter_handle, ICARSAMODEXPO, &rb);
+		if (any_card_online)
+			rc = ioctl(adapter_handle, ICARSAMODEXPO, &rb);
+		else
+			rc = ENODEV;
+
 		if (!rc)
 			hardware = ALGO_HW;
 		else
@@ -1126,7 +1130,11 @@ unsigned int ica_rsa_crt(ica_adapter_handle_t adapter_handle,
 		rc = ica_fallbacks_enabled ?
 			rsa_crt_sw(&rb) : ENODEV;
 	else {
-		rc = ioctl(adapter_handle, ICARSACRT, &rb);
+		if (any_card_online)
+			rc = ioctl(adapter_handle, ICARSACRT, &rb);
+		else
+			rc = ENODEV;
+
 		if(!rc)
 			hardware = ALGO_HW;
 		else
--- a/src/s390_crypto.c
+++ b/src/s390_crypto.c
@@ -5,8 +5,8 @@
  */
 
 /*
- * Entry point for crypto graphic instructions on s390. If a instruction
- * is not available, instruction will be simulated in software.
+ * Entry point for crypto graphic instructions on s390. If an instruction
+ * is not available, the related function is simulated in software.
  *
  * Authors(s): Ralph Wuerthner <rwuerthn@de.ibm.com>
  *	       Jan Glauber <jan.glauber@de.ibm.com>
@@ -332,14 +332,14 @@ void s390_crypto_switches_init(void)
 /*
  * The first field represents the mechanism ID.
  * The second field represents the function family type (category),
- * The third filed represents the function code.
+ * The third field represents the function code.
  * This function code will be used later to check if HW support
  * is available and modifies the SW/HW-support-flag.
  * SHW - static hardware support (CPACF)
  * DHW - dynamic hardware support (crypto adapter)
  * SW  - software support
  * Bit field flags: [0|0|0|0|0|SHW|DHW|SW]
- * The last filed represent the property flags
+ * The last field represents the property flags indicating key lengths
  */
 libica_func_list_element_int icaList[] = {
  {SHA1,   KIMD, SHA_1  , ICA_FLAG_SW, 0},
--- a/src/s390_crypto.c
+++ b/src/s390_crypto.c
@@ -406,76 +406,76 @@ libica_func_list_element_int icaList[] = {
  * Query s390_xxx_functions for each algorithm to check
  * CPACF support and update the corresponding SHW-flags.
  */
-int s390_initialize_functionlist() {
-
-  unsigned int list_len = sizeof(icaList)/sizeof(libica_func_list_element_int);
-
-  unsigned int x;
-  for (x = 0; x < list_len; x++) {
-	libica_func_list_element_int *e = &icaList[x];
-	switch ((int) e->type) {
-	case KIMD:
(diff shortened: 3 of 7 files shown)


src/s390_crypto.c
```diff
@@ -467,8 +574,19 @@ int s390_initialize_functionlist()
...
 		case EC_KGEN:
+			if (ecc_via_online_card) {
+				e->flags |= ICA_FLAG_DHW;
+				e->property = 0x0F;
+			}
 			e->flags |= *s390_kdsa_functions[e->id].enabled ? ICA_FLAG_SHW : 0;
 			break;
+		case RSA_ME: /* fall-through */
+		case RSA_CRT:
+			if (any_card_online) {
```
p-steuer: I think libica RSA only works with CEX-A, no ? So 'any_card_online' would be neccesary but not sufficient.

jschmidb: I found this paragraph in the "Device Drivers, Features, and Commands" book:

> For both accelerators and CCA coprocessors, it provides Rivest-Shamir-Adleman
> (RSA) encryption and RSA decryption functions using clear keys. RSA operations
> are supported in both the modulus-exponent and the Chinese-Remainder Theorem
> (CRT) variants for any key size in the range 57 - 4096 bit.

So it looks like the zcrypt device driver can use both types of cards for RSA.

jschmidb: Just tested it: the RSA tests are performed in hw on a CCA coprocessor.

include/ica_api.h
```diff
@@ -68,10 +68,38 @@ typedef ica_adapter_handle_t ICA_ADAPTER_HANDLE;
...
+#define ICA_PROPERTY_RSA_4096		0x00000008
+
+#define ICA_PROPERTY_RSA_ALL	\
+	(ICA_PROPERTY_RSA_512 |		\
+	 ICA_PROPERTY_RSA_1024 |	\
+	 ICA_PROPERTY_RSA_2048 |	\
+	 ICA_PROPERTY_RSA_4096)
+
+#define ICA_PROPERTY_EC_BP			0x00000001 /* Brainpool curves */
+#define ICA_PROPERTY_EC_NIST		0x00000002 /* NIST curves */
```
ifranzki: Is there a whitespace error ? The NIST curves constant is indented differently than the others in the GitHub diff view....

src/s390_crypto.c
```diff
@@ -33,7 +34,13 @@ unsigned long long facility_bits[3];
 unsigned int sha1_switch, sha256_switch, sha512_switch, sha3_switch, des_switch,
 	     tdes_switch, aes128_switch, aes192_switch, aes256_switch,
 	     prng_switch, tdea128_switch, tdea192_switch, sha512_drng_switch,
-	     msa4_switch, msa5_switch, msa8_switch, trng_switch, msa9_switch;
+	     msa4_switch, msa5_switch, msa8_switch, trng_switch, msa9_switch,
+		 ecc_via_online_card, any_card_online;
+
+#define CARD_AVAILABLE		0x01
```
ifranzki: also here it looks like a whitespace error......

src/s390_crypto.c
```diff
@@ -318,14 +325,114 @@ static void set_switches(int msa)
...
+	file = fopen(dev, "r");
+	if (file == NULL)
+		return 0;
+
+	/* Read device type: e.g. 'CEX5C' */
+	if (getline(&type, &size, file) == -1)
+		goto end;
+
+	/* Did we get something like 'CEX5C\n'? */
+	if (strlen(type) != 6 || strstr(type, "CEX") != &type[0])
```
ifranzki: didn't we say that strstr might not be the best choice here? I would do a `strncmp(type, "CEX", 3)` instead.

src/s390_crypto.c
```diff
@@ -318,14 +325,114 @@ static void set_switches(int msa)
...
+	struct dirent *direntp;
+	char type[6];
+
+	if ((sysDir = opendir(dev)) == NULL)
+		return 0;
+
+	while ((direntp = readdir(sysDir)) != NULL) {
+
+		/* Skip entries that are not like "card01", "card02", etc. */
+		if (strstr(direntp->d_name, "card") != &direntp->d_name[0])
```
ifranzki: `strncmp(... ,"card", 4) `

src/s390_crypto.c
```diff
@@ -318,14 +325,114 @@ static void set_switches(int msa)
...
+
+end:
+	free(type);
+	fclose(file);
+
+	return ret;
+}
+
+#define AP_PATH  "/sys/devices/ap"
+#define MAX_DEV_LEN 280
```
hfreude: There is a PATH_MAX define in linux/limits.h.

src/s390_crypto.c
```diff
@@ -318,14 +325,114 @@ static void set_switches(int msa)
...
+
+unsigned int search_for_cards()
+{
+	DIR *sysDir;
+	unsigned int ret = 0;
+	char dev[MAX_DEV_LEN] = AP_PATH;
```
hfreude: Why the = AP_PATH ?
You use dev later with snprintf(dev, ...) so the assignment is not needed.

src/s390_crypto.c
```diff
@@ -318,14 +325,114 @@ static void set_switches(int msa)
...
+			continue;
+
+		/* Check if device online */
+		snprintf(dev, MAX_DEV_LEN, "%s/%s/online", AP_PATH, direntp->d_name);
+		if (!is_device_online(dev))
+			continue;
+
+		/* Get device type (string like "CEXnT") */
+		snprintf(dev, MAX_DEV_LEN, "%s/%s/type", AP_PATH, direntp->d_name);
+		memset(type, 0, sizeof(type));
```
hfreude: the memset here is not needed as you check the returncode from get_device_type()

src/s390_crypto.c
hfreude: Yes, the zcrypt dd can use both.

include/ica_api.h
```diff
@@ -68,10 +68,38 @@ typedef ica_adapter_handle_t ICA_ADAPTER_HANDLE;
...
+ * ICA properties: key lengths
+ */
+#define ICA_PROPERTY_AES_128		0x00000001
+#define ICA_PROPERTY_AES_192		0x00000002
+#define ICA_PROPERTY_AES_256		0x00000004
+
+#define ICA_PROPERTY_RSA_512		0x00000001
+#define ICA_PROPERTY_RSA_1024		0x00000002
+#define ICA_PROPERTY_RSA_2048		0x00000004
+#define ICA_PROPERTY_RSA_4096		0x00000008
```
hfreude: do we really need this differentiation ? I know there was a limit on RSA for CEX2 but as far as I know even CEX3 supports RSA up to 4096 and we do not have any CEX2 support in the device driver any more.

src/s390_crypto.c
jschmidb: Yes, but it's only 255 bytes long and causes a compiler warning:

> s390_crypto.c:401:34: warning: '%s' directive output may be truncated writing up to 255 bytes into a region of size 239 [-Wformat-truncation=]
>   401 |   snprintf(dev, MAX_DEV_LEN, "%s/%s/type", AP_PATH, direntp->d_name);

The warning disappears when increasing the array to 280 bytes.

include/ica_api.h
jschmidb: Yes, you are right, today we don't really need it. We could just use ICA_PROPERTY_RSA_ALL. I left the differentiation, because the libica book describes the different RSA key lengths in chapter 5. We may remove the differentiation in the code, but should then also update the book.

src/s390_crypto.c
jschmidb: Yes, that's right.

src/s390_crypto.c
jschmidb: Yes, but there's this stmt below which needs the path in dev:

> 	if ((sysDir = opendir(dev)) == NULL)
> 		return 0;

include/ica_api.h
jschmidb: I will fix this.

src/s390_crypto.c
jschmidb: will fix it.

src/s390_crypto.c
jschmidb: ok

src/s390_crypto.c
jschmidb: ok

src/ica_api.c
```diff
@@ -1017,7 +1017,11 @@ unsigned int ica_rsa_mod_expo(ica_adapter_handle_t adapter_handle,
 		rc = ica_fallbacks_enabled ?
 			rsa_mod_expo_sw(&rb) : ENODEV;
 	else {
-		rc = ioctl(adapter_handle, ICARSAMODEXPO, &rb);
+		if (any_card_online)
+			rc = ioctl(adapter_handle, ICARSAMODEXPO, &rb);
+		else
+			rc = 1;
```
ifranzki: Is there a reason why you return "1" here, an not ENODEV or similar?

src/ica_api.c
```diff
@@ -1126,7 +1130,11 @@ unsigned int ica_rsa_crt(ica_adapter_handle_t adapter_handle,
 		rc = ica_fallbacks_enabled ?
 			rsa_crt_sw(&rb) : ENODEV;
 	else {
-		rc = ioctl(adapter_handle, ICARSACRT, &rb);
+		if (any_card_online)
+			rc = ioctl(adapter_handle, ICARSACRT, &rb);
+		else
+			rc = 1;
```
ifranzki: same here

src/ica_api.c
jschmidb: I agree, ENODEV is better here. Code changed.

src/ica_api.c
jschmidb: I agree, ENODEV is better here. Code changed.

PR #30, opened 2019-08-30
Add missing -pthread flag.

xnox:
Signed-off-by: Dimitri John Ledkov <xnox@ubuntu.com>


p-steuer:
better add -pthread this to compiler flags, no ?

xnox:
Doesn't it actually need to be in both, or at least in ldflags to trigger linking with pthreads, no?!

p-steuer:
I agree it should be used in both linking, compiling and preprocessing.

My understanding is that when using gcc with its pthread option (see gcc(3)), the flag will be used for preprocesser, compiler and linker.

p-steuer:
If i add it to CFLAGS, the flag is used with libtools link and compile mode.
So i ll merge https://github.com/opencryptoki/libica/pull/32 instead of this one if there are no objections.

p-steuer:
PR https://github.com/opencryptoki/libica/pull/32 was merged. closing this one.

PR #39, opened 2020-03-18
Fix DES and TDES key length.

jschmidb:
Although a DES key has only 56 effective bits, all 64 bits must be
considered, because the parity bits are spread over all 8 bytes of
the key.

### DIFF
--- a/src/ica_api.c
+++ b/src/ica_api.c
@@ -48,6 +48,8 @@
 #define DEFAULT2_CRYPT_DEVICE "/dev/z90crypt"
 #define DEFAULT3_CRYPT_DEVICE "/dev/zcrypt"
 
+#define DES_KEY_LEN64		(64/8)
+
 #define MAX_VERSION_LENGTH 16
 
 int ica_fallbacks_enabled = 1;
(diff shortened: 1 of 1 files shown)


ifranzki:
OK, DES_KEY_LENGTH is only used in fips_check_3des_key(), and DES3_KEY_LENGTH is not used anywhere. So from that point of view it looks OK.

The potential problem I see is that ica_api.h is an external header file, included into user applications. Not sure what these applications do with the 2 defines. It might cause problems when the 2 defines are now changed....

Maybe it would be saver to add 2 new defines inside ica_api.c or into an internal header file, and use them in fips_check_3des_key(), but leave the 2 external defines as is to not break anything.

p-steuer:
I agree with ingo.

jschmidb:
Good point Ingo ... we could also add two new constants in ica_api.h so that everyone sees both constants, and users could use both constants in their apps. Like:

> #define DES_KEY_LENGTH		(56/8)
> #define DES3_KEY_LENGTH	(168/8)
> #define DES_KEY_SIZE		        (64/8)
> #define DES3_KEY_SIZE		(192/8)
We would then use the new constants for the FIPS key checks.

p-steuer:
Please dont add more constants to the api, it would be quite confusing to define both KEY_LEGNTH and KEY_SIZE. Keep those additional constants internal, as ingo suggested.

jschmidb:
OK, I added one new (internal) constant in ica_api.c and the fips_check_3des_key() now uses this new constant. A second new constant for TDES would not be used, same as DES3_KEY_LENGTH today, so let's add this when we really need it.

PR #40, opened 2020-03-25
FIPS: provide output iv as required by FIPS tests

jschmidb:
This fix is introduced to satisfy FIPS tests. They require the
output iv to be the iv resulting from decrypting the last block
with a zero iv as input. But note that this is not described
in the NIST standard for CBC-CS. According to the standard,
the output iv is simply undefined.


p-steuer:
Do you want to add the corresponding test vector to the fips startup tests ?

jschmidb:
I have one test vector in the Bugzilla, consisting of a key, some ciphertext, and the expected output iv. This doesn't really fit into the existing cbc-cs tests. They do encrypt/decrypt and never check the output iv. So I can either add this test separately, or check with the originator of the Bigzilla, if they can provide me a complete test vector.

PR #41, opened 2020-03-27
icainfo: bugfix for RSA and EC related info for software column.

jschmidb:
The software column did never show 'yes' for RSA ME, RSA CRT, ECDH,
ECDSA sign, ECDSA verify, and ECKGEN, due to the missing ICA_FLAG_SW
flag in the icaList.


ifranzki:
For my understanding: So without these flags, this caused rc=0 on those operations, when no HW was available, and the software fallback was also not taken? If so, then maybe you should also fix that rc=0 is not returned in such cases.

jschmidb:
No, the sw fallback is only dependent on an internal flag ica_fallbacks_enabled, which is 1 by default, but can be changed by an application via an API function. This commit fixes icainfo only. 

An rc = 0 from ica_rsa_crt without hw available and the sw path not taken is not possible in my opinion. But I will investigate this issue separately. I could not reproduce this error in my test environment, but we will see ...

PR #42, opened 2020-04-27
FIPS: introduce HMAC based library integrity check

jschmidb:
When in FIPS mode, perform an integrity check on libica.so by calculating
an HMAC from the file contents using a static HMAC key, and comparing it
to a pre-calculated HMAC in a separate file. The HMAC key and HMAC file
may be provided by a Distributor when building the packet. The test
succeeds if the HMAC file is not present.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -9,7 +9,7 @@ lib_LTLIBRARIES = libica.la
 libica_la_CFLAGS = ${AM_CFLAGS} -I${srcdir}/include -I${srcdir}/../include \
 		   -fvisibility=hidden -pthread
 libica_la_CCASFLAGS = ${AM_CFLAGS}
-libica_la_LIBADD = @LIBS@ -lrt -lcrypto
+libica_la_LIBADD = @LIBS@ -lrt -lcrypto -ldl
 libica_la_LDFLAGS = -Wl,--version-script=${srcdir}/../libica.map \
 		    -version-number ${VERSION}
 libica_la_SOURCES = ica_api.c init.c icastats_shared.c s390_rsa.c \
@@ -53,7 +53,7 @@ internal_tests_ec_internal_test_CFLAGS = ${AM_CFLAGS} -I${srcdir}/include \
 					 -DICA_INTERNAL_TEST \
 					 -DICA_INTERNAL_TEST_EC
 internal_tests_ec_internal_test_CCASFLAGS = ${AM_CFLAGS}
-internal_tests_ec_internal_test_LDADD = @LIBS@ -lrt -lcrypto -lpthread
+internal_tests_ec_internal_test_LDADD = @LIBS@ -lrt -lcrypto -lpthread -ldl
 internal_tests_ec_internal_test_SOURCES = \
 		    ica_api.c init.c icastats_shared.c s390_rsa.c \
 		    s390_crypto.c s390_ecc.c s390_prng.c s390_sha.c \
--- a/src/fips.c
+++ b/src/fips.c
@@ -13,11 +13,13 @@
 
 #include <errno.h>
 #include <openssl/crypto.h>
+#include <openssl/evp.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include <syslog.h>
+#include <dlfcn.h>
 
 #include <openssl/opensslconf.h>
 #ifdef OPENSSL_FIPS
@@ -27,6 +29,24 @@
 #include "fips.h"
 #include "ica_api.h"
 #include "test_vec.h"
+
+#ifndef PATH_MAX
+#define PATH_MAX 4096
+#endif
+
+#define HMAC_PREFIX "."
+#define HMAC_SUFFIX ".hmac"
+#define READ_BUFFER_LENGTH 16384
+
+/*
+ * The hard-coded HMAC key to be optionally provided for the library
+ * integrity test. The recommended key size for HMAC-SHA256 is 64 bytes.
+ * The known HMAC is supposed to be provided as hex string in a file
+ * libica.so.MAJOR.hmac in the same directory as the .so module.
+ */
+static const char hmackey[] =
+	"0000000000000000000000000000000000000000000000000000000000000000"
+	"0000000000000000000000000000000000000000000000000000000000000000";
 
 int fips;
 
@@ -101,6 +121,206 @@ fips_init(void)
 		FIPS_mode_set(1);
 	}
 }
+static int get_library_path(const char *libname, const char *symbolname,
+							char *path, size_t pathlen)
+{
+	Dl_info info;
+	void *dl, *sym;
+	int rc = -1;
+
+	dl = dlopen(libname, RTLD_LAZY);
+	if (dl == NULL)
+		goto done;
+
+	sym = dlsym(dl, symbolname);
+	if (sym != NULL && dladdr(sym, &info)) {
+		if (strlen(info.dli_fname) < pathlen)
+			strcpy(path, info.dli_fname);
+		else
+			goto done;
+	}
+
+	rc = 0;
+
+done:
+	if (dl != NULL)
+		dlclose(dl);
+
+	return rc;
+}
+
+static char *make_hmac_path(const char *origpath)
+{
+	char *path;
+	const char *fn;
+
+	path = malloc(sizeof(HMAC_PREFIX) + sizeof(HMAC_SUFFIX) + strlen(origpath) + 1);
+	if (path == NULL)
+		return NULL;
+
+	fn = strrchr(origpath, '/');
+	if (fn == NULL) {
+		fn = origpath;
+	} else {
+		++fn;
+	}
+
+	strncpy(path, origpath, fn - origpath);
+	strcat(path, HMAC_PREFIX);
+	strcat(path, fn);
+	strcat(path, HMAC_SUFFIX);
+
+	return path;
+}
+
+static int compute_file_hmac(const char *path, void **buf, size_t *hmaclen)
+{
+	FILE *fp = NULL;
+	int rc = -1;
+	unsigned char rbuf[READ_BUFFER_LENGTH];
+	unsigned char *keybuf;
+	EVP_MD_CTX *mdctx = NULL;
+	EVP_PKEY *pkey = NULL;
+	size_t hlen, len;
+	long keylen;
+
+	keybuf = OPENSSL_hexstr2buf(hmackey, &keylen);
+	pkey = EVP_PKEY_new_mac_key(EVP_PKEY_HMAC, NULL, keybuf, (int)keylen);
+	if (!pkey)
+		goto end;
+
+	mdctx = EVP_MD_CTX_create();
+	if (!mdctx)
+		goto end;
+
+	fp = fopen(path, "r");
+	if (fp == NULL)
+		goto end;
+
+	if (EVP_DigestSignInit(mdctx, NULL, EVP_sha256(), NULL, pkey) <= 0)
+		goto end;
+
+	while ((len = fread(rbuf, 1, sizeof(rbuf), fp)) != 0) {
+		if (EVP_DigestSignUpdate(mdctx, rbuf, len) <= 0) {
+			goto end;
+		}
+	}
+
+	if (EVP_DigestSignFinal(mdctx, rbuf, &hlen) <= 0)
+		goto end;
+
+	*buf = malloc(hlen);
+	if (*buf == NULL)
+		goto end;
+
+	*hmaclen = hlen;
+
+	memcpy(*buf, rbuf, hlen);
+
+	rc = 0;
+
+end:
+
+	if (pkey != NULL)
+		EVP_PKEY_free(pkey);
+
+	free(keybuf);
+	EVP_MD_CTX_destroy(mdctx);
+	if (fp)
+		fclose(fp);
+
+	return rc;
+}
+
+/**
+ * Performs the FIPS check.
+ *
+ * @return  1 if check succeeded
+ *          0 otherwise
+ */
+static int FIPSCHECK_verify(const char *path)
+{
+	int rc = 0;
+	FILE *fp;
+	unsigned char *hmac_buf = NULL;
+	long hmaclen;
+	char *hmacpath, *p;
+	char *hmac_str = NULL;
+	size_t n, buflen;
+	void *buf = NULL;
+
+	hmacpath = make_hmac_path(path);
+	if (hmacpath == NULL)
+		return 0;
+
+	fp = fopen(hmacpath, "r");
+	if (fp == NULL) {
+		rc = 1;
+		goto end;
+	}
+
+	if (getline(&hmac_str, &n, fp) <= 0)
+		goto end;
+
+	if ((p = strchr(hmac_str, '\n')) != NULL)
+		*p = '\0';
+
+	hmac_buf = OPENSSL_hexstr2buf(hmac_str, &hmaclen);
+
+	if (compute_file_hmac(path, &buf, &buflen) != 0)
+		goto end;
+
+	if (memcmp(buf, hmac_buf, hmaclen) != 0)
+		goto end;
+
+	rc = 1;
+
+end:
+
+	free(buf);
+	free(hmac_str);
+	free(hmacpath);
+
+	OPENSSL_free(hmac_buf);
+
+	if (fp)
+		fclose(fp);
+
+	return rc;
+}
+
+static const char msg1[] = "Libica FIPS library integrity check failed. Cannot determine library path.\n";
+static const char msg2[] = "Libica FIPS library integrity check failed. Module %s probably corrupted.\n";
+static const char msg3[] = "Libica FIPS library integrity check passed.\n";
+
+/*
+ * Perform an integrity check on libica.so by calculating an HMAC from
+ * the file contents using a static HMAC key, and comparing it to a
+ * pre-calculated HMAC in a separate file. The HMAC key and HMAC file
+ * may be provided by a Distributor when building the packet.
+ */
+static void fips_lib_integrity_check(void)
+{
+	int rc;
+	char path[PATH_MAX];
+	const char *libname = "libica.so";
+	const char *symbolname = "ica_sha256";
+
+	rc = get_library_path(libname, symbolname, path, sizeof(path));
+	if (rc != 0) {
+		syslog(LOG_ERR, msg1);
+		fips |= ICA_FIPS_INTEGRITY;
+		return;
+	}
+
+	if (!FIPSCHECK_verify(path)) {
+		syslog(LOG_ERR, msg2, path);
+		fips |= ICA_FIPS_INTEGRITY;
+		return;
+	}
+
+	syslog(LOG_INFO, msg3);
+}
 
 void
 fips_powerup_tests(void)
@@ -117,6 +337,9 @@ fips_powerup_tests(void)
 		fips |= ICA_FIPS_CRYPTOALG;
 		return;
 	}
+
+	/* Library integrity test */
+	fips_lib_integrity_check();
 }
 
 static int
@@ -933,5 +1156,4 @@ _err_:
 	syslog(LOG_ERR, "Libica RSA test failed.");
 	return 1;
 }
-
 #endif /* FIPS_H */
--- a/src/include/fips.h
+++ b/src/include/fips.h
@@ -27,8 +27,8 @@ extern int fips;			/* module status */
 void fips_init(void);
 
 /*
- * Powerup tests: crypto algorithm test, SW/FW integrity test (not implemented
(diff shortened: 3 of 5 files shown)


src/fips_check.c
```diff
@@ -0,0 +1,224 @@
...
+
+	sym = dlsym(dl, symbolname);
+	if (sym != NULL && dladdr(sym, &info)) {
+		if (strlen(info.dli_fname) <= pathlen)
+			strcpy(path, info.dli_fname);
+		else
+			goto done;
+	}
+
+	dlclose(dl);
```
p-steuer, line 55: move this below the "done" label with a dl != NULL check.

src/fips_check.c
```diff
@@ -0,0 +1,224 @@
...
+	void *dl, *sym;
+	int rc = -1;
+
+	dl = dlopen(libname, RTLD_LAZY);
+	if (dl == NULL)
+		goto done;
+
+	sym = dlsym(dl, symbolname);
+	if (sym != NULL && dladdr(sym, &info)) {
+		if (strlen(info.dli_fname) <= pathlen)
```
p-steuer, line 49: " < pathlen", pathlen is the size of the buffer and needs 1 byte for the terminating '\0'.

src/fips_check.c
```diff
@@ -0,0 +1,224 @@
...
+#include <syslog.h>
+#include <dlfcn.h>
+
+
+#ifndef PATH_MAX
+#define PATH_MAX 4096
+#endif
+
+#define HMAC_PREFIX "."
+#ifndef HMAC_SUFFIX
```
p-steuer, line 29: Why is this ifndef needed ? If there are possible collissions with the includes, use an ICA_ prefix.

src/fips_check.c
```diff
@@ -0,0 +1,224 @@
...
+
+	c = HMAC_CTX_new();
+	if (c == NULL)
+		return rc;
+
+	fp = fopen(path, "r");
+	if (fp == NULL)
+		goto end;
+
+	if (HMAC_Init_ex(c, hmackey, sizeof(hmackey) - 1, EVP_sha256(), NULL) <= 0)
```
p-steuer, line 105: HMAC_ interfaces will be deprecated with openssl 3.0. Can we use the EVP_ interface instead ?

p-steuer: E.g., look at opencryptoki's softtok_hmac_init in soft_specific.c

src/fips_check.c
```diff
@@ -0,0 +1,224 @@
...
+#define PATH_MAX 4096
+#endif
+
+#define HMAC_PREFIX "."
+#ifndef HMAC_SUFFIX
+#define HMAC_SUFFIX ".hmac"
+#endif
+#define READ_BUFFER_LENGTH 16384
+
+static const char hmackey[] = "00000000000000000000000000000000";
```
p-steuer, line 34: do we really want to encode the key in ascii ? ascii 0 == 48. I think we should use static cons tunsigned char hmacckey[32] = {0} i.e. the all-zero-key, instead of an all-48-key.

src/Makefile.am
```diff
@@ -9,13 +9,13 @@ lib_LTLIBRARIES = libica.la
...
 libica_la_CCASFLAGS = ${AM_CFLAGS}
-libica_la_LIBADD = @LIBS@ -lrt -lcrypto
+libica_la_LIBADD = @LIBS@ -lrt -lcrypto -ldl
 libica_la_LDFLAGS = -Wl,--version-script=${srcdir}/../libica.map \
 		    -version-number ${VERSION}
 libica_la_SOURCES = ica_api.c init.c icastats_shared.c s390_rsa.c \
 		    s390_crypto.c s390_ecc.c s390_prng.c s390_sha.c \
 		    s390_drbg.c s390_drbg_sha512.c test_vec.c fips.c \
-		    mp.S rng.c \
+		    mp.S rng.c fips_check.c \
```
p-steuer, line 18: we already have a fips.c, any reason why the fips intergrity check needs a seperate source file ?

src/fips_check.c
```diff
@@ -0,0 +1,224 @@
...
+
+	return rc;
+}
+
+static char *make_hmac_path(const char *origpath)
+{
+	char *path, *p;
+	const char *fn;
+
+	path = malloc(sizeof(HMAC_PREFIX) + sizeof(HMAC_SUFFIX) + strlen(origpath));
```
p-steuer, line 68: +1 byte for terminating '\0' ?

src/fips_check.c
```diff
@@ -0,0 +1,224 @@
...
+
+	fn = strrchr(origpath, '/');
+	if (fn == NULL) {
+		fn = origpath;
+	} else {
+		++fn;
+	}
+
+	strncpy(path, origpath, fn - origpath);
+	p = path + (fn - origpath);
```
p-steuer, line 80: why not use strcat to concat the strings and get rid of tracking the string's final '\0' byte with p ?

src/fips_check.c
```diff
@@ -0,0 +1,224 @@
...
+
+	if (memcmp(buf, hmac_buf, hmaclen) != 0)
+		goto end;
+
+	rc = 1;
+
+end:
+
+	if (buf)
+		free(buf);
```
p-steuer, line 184: free already does the NULL checking.

src/include/fips.h
```diff
@@ -20,6 +20,14 @@
 extern int fips;			/* module status */
 
 
+/*
+ * Perform an integrity check on libica.so by calculating an HMAC from
+ * the file contents using a static HMAC key, and comparing it to a
+ * pre-calculated HMAC in a separate file. The HMAC key and HMAC file
+ * may be provided by a Distributor when building the packet.
+ */
+void fips_lib_integrity_check(void);
```
p-steuer, line 29: since you put the prototype in fips.h, id move the corresponding function to fips.c (from fips_check.c).

src/init.c
```diff
@@ -93,6 +95,8 @@ void __attribute__ ((constructor)) icainit(void)
 		ica_set_stats_mode(value);
 
 #ifdef ICA_FIPS
+	static CRYPTO_ONCE once_run = CRYPTO_ONCE_STATIC_INIT;
+	CRYPTO_THREAD_run_once(&once_run, fips_lib_integrity_check);
```
p-steuer, line 99: we are in the constructor here. any reason to use libcrypto's run_once ? and why did you put it before fips_init ? Did you check if fips imposes an order on the start up checks ?

p-steuer: I think the integrity check is part of the power up test and should be called from fips_powerup_tests (fips.c). If it fails it should set the ICA_FIPS_INTEGRITY (ica_api.h) in the global fips module status flags (fips variable in fips.c). That is how it was supposed to work.

src/fips_check.c
```diff
@@ -0,0 +1,224 @@
...
+	size_t n, buflen;
+	void *buf = NULL;
+
+	hmacpath = make_hmac_path(path);
+	if (hmacpath == NULL)
+		return 0;
+
+	fp = fopen(hmacpath, "r");
+	if (fp == NULL) {
+		rc = 1;
```
hfreude, line 161: hm, add some kind of logging / syslog entry here ?

src/fips_check.c
```diff
@@ -0,0 +1,224 @@
...
+
+	hmac_buf = OPENSSL_hexstr2buf(hmac_str, &hmaclen);
+
+	if (compute_file_hmac(path, &buf, &buflen) != 0)
+		goto end;
+
+	if (memcmp(buf, hmac_buf, hmaclen) != 0)
+		goto end;
+
+	rc = 1;
```
hfreude, line 179: and also for the successful check, add some kind of logging / syslog here ?

src/fips_check.c
```diff
@@ -0,0 +1,224 @@
...
+
+static const char msg1[] = "Libica FIPS library integrity check failed. Cannot determine library path.\n";
+static const char msg2[] = "Libica FIPS library integrity check failed. Module %s probably corrupted.\n";
+static const char msg3[] = "Libica FIPS library integrity check passed.\n";
+
+void fips_lib_integrity_check(void)
+{
+	int rc;
+	char path[PATH_MAX];
+	char *libname = "libica.so";
```
hfreude, line 206: const char *libname ...

src/fips_check.c
```diff
@@ -0,0 +1,224 @@
...
+	char *symbolname = "ica_sha256";
```
hfreude, line 207: const char *symbolname ...

src/fips_check.c
hfreude: forget it, the calling function does this

src/fips_check.c
hfreude: As this is 'silently' assumed to be a full check and reported as 'successful' checked, I'd like to see either an additional syslog entry or you could instead return some kind of treestate: -1 for failure, 0 for not checked because of missing hmac value file, 1 for successfuly checked

src/fips_check.c
hfreude, line 206: Do we need this a little bit more high sophisticated ?
For example:
- check if libica.so is a softlink
- if yes, find the source of the softlink
- again check for softlink until finally the real so file is found.
- Now use the real so file name and the derived hmac file for the checks.
The advantage would be to able to hold more than one libica.so.xxx with all the related libica.so.xxx.hmac files.

src/fips_check.c
hfreude, line 34: I'd like to see a comment here how this hmackey is actually filled with the real value.

src/fips_check.c
jschmidb: Done.

src/fips_check.c
jschmidb: Done.

src/fips_check.c
jschmidb: Removed.

src/fips_check.c
jschmidb: As our distributor will encode the known HMAC in the .hmac file in the same way as string, I would prefer to leave it as a string here as well. But you are right about the init value, which is misleading. What about just
> static const char hmackey[32];

without an init value. However, it's more obvious for Distributors how to insert the known HMAC key when we leave it as it is. To really provide this as a placeholder, we could even do it like:
>  static const char hmackey[] = "000102030405060708090A0B0C0D0E0F";

And yes, this would be only 16 bytes ... maybe we should then show an example with 32 bytes.

src/Makefile.am
jschmidb: Yes, because this integrity check comprises 5 routines and a couple of definitions, I put it into a separate file. But there is no other reason why not adding the code to fips.c

src/fips_check.c
jschmidb: Done.

src/fips_check.c
jschmidb: Done.

src/include/fips.h
jschmidb: see answer above: no problem to add all code from fips_check.c to fips.c, or just only fips_lib_integrity_check. However, if we decide to move fips_lib_integrity_chec to fips.c then I would also move the rest.

src/init.c
jschmidb: ok, you're right,  the run_once is superfluous. 
Re: order on the fips start up checks: I don't think there is an order. If running in fips mode and any of the tests fails, the lib should terminate. But we may move the integrity check after the power up test. 
Re: ICA_FIPS_INTEGRITY: I wasn't aware of this flag. OK, I will handle this also in fips_test.c similar to ICA_FIPS_CRYPTOALG.
But what about the exit(1); in fips_lib_integrity_check? I think the lib must exit if this check fails.

src/fips_check.c
jschmidb: see below ...

src/fips_check.c
jschmidb: yep.

src/fips_check.c
jschmidb: Done.

src/fips_check.c
jschmidb: Done.

src/fips_check.c
p-steuer: Okay, but if you encode the hmac key as a string (like the known mac in the file), you should process it with OPENSSL_hexstr2buf before using it (as you do for the known mac). In that case it would be okay to initialize hmackey to a string of "0...". Id use a 32 byte all zero key as default value.

src/fips_check.c
jschmidb: Good idea! But let's add this in a separate step.

src/fips_check.c
jschmidb: Done.

src/Makefile.am
p-steuer: Id put everything in fips.c/.h

src/include/fips.h
p-steuer: Id put everything in fips.c/.h

src/init.c
p-steuer: See FIPS 140-2 chap. 4.9: 

Self-tests:
- power-up test (algorithms, integrity, critical functions)
- conditional tests

So the integrity test is part of the power up tests (and thus should be called from libica's fips_powerup_tests). The spec doesnt impose an order of the power-up tests, but id just use the order in which the specs lists the tests i.e., but it below the cryptographic algorithm tests in the power up test function. (the power up test's critical function test are n/a for libica).

Regarding exit: the spec says:
```
If a cryptographic module fails a self-test, the module shall enter an error state and output an error indicator via the status output interface.  The cryptographic module shall not perform any cryptographic operationswhile  in  an  error  state.  All  data  output  via  the  data  output  interface  shall  be  inhibited  when  an  error  state  exists. 
```

So right now, in case of fips error state, all libica functions will return a corresponding error. Do you think this is not enough? Why do you think the library, resp. the application which links the library, should always exit ?

src/fips_check.c
ifranzki: Don't we have versioned so names nowadays? E.g. libica.so.3...
I would make the so name a define, so that it can be easily changed, and possibly patched by the distributor when they build versioned so names.

src/init.c
ifranzki: Hm, not sure if a library should ever do exit(). This could end a process silently, and nobody knows why. So if you need to exit the hard way, then use abort(), this will at least be seen by the user.
I am still not sure if a library should abort, or influence the life of the calling process in any way. Maybe that's what we want for the FIPS checks though....

src/fips_check.c
p-steuer: What i mean is: afaict right now, the values of the key's bytes are the values from the ascii table for the corresponding letter in the string but i think it should be the hex value encoded by two letters of the string.

src/fips_check.c
jschmidb: Done.

src/fips_check.c
jschmidb: Done.

src/fips_check.c
jschmidb: OK, now using OPENSSL_hexstr2buf. I also noticed that the recommended key size for HMAC-SHA256 is even 64 bytes. So I'm now providing a "zero"-key with 64 bytes:
> static const char hmackey[] =
> 	"0000000000000000000000000000000000000000000000000000000000000000"
> 	"0000000000000000000000000000000000000000000000000000000000000000";

Distributors are however free to change this.

src/Makefile.am
jschmidb: Done.

src/include/fips.h
jschmidb: Done.

src/init.c
jschmidb: OK, doing the exit() was the natural way for me to handle this. But when the spec says we should go into an error state this is also fine. It now just sets the ICA_FIPS_INTEGRITY flag. I also moved the integrity check to the power up test.

src/fips_check.c
jschmidb: Yes, but the get_library_path routine already returns the correct path including major number, e.g. "libica.so.3". And the make_hmac_path routine produces the correct corresponding pathname of the .hmac file, e.g. /usr/lib/.libica.so.3.hmac. So there's no need to specify the major number somewhere in the code.

src/init.c
jschmidb: One last question: Do we want an additional fprintf when the check fails?
```
	if (!FIPSCHECK_verify(path)) {
		syslog(LOG_ERR, msg2, path);
		fprintf(stderr, msg2, path);
		fips |= ICA_FIPS_INTEGRITY;
	}
```
When the check fails, we get a 
> FIPS SELF-TEST FAILURE. CHECK THE SYSLOG.

So an additional message like
> Libica FIPS library integrity check failed. Module /usr/local/lib/libica.so.3 probably corrupted.

is not really necessary. Other fips tests don't issue extra messages.
What do you think?

p-steuer: > Do we want an additional fprintf when the check fails?

No, the library should not print to stderr/out.

> FIPS SELF-TEST FAILURE. CHECK THE SYSLOG.

This message is displayed by the icainfo tool, not printed by the library on failure.

A syslog entry for the integrity check failure is enough, in case other fips sleft tests fail, we do the same e.g.:
> syslog(LOG_ERR, "Libica internal RNG error..");

src/include/fips.h
p-steuer, line 29: fips_lib_integrity_check is only used in fips.c, so you can remove the prototype from fips.h and make it static in fips.c.

jschmidb: Done.

src/fips.c
```diff
@@ -28,6 +30,24 @@
 #include "ica_api.h"
 #include "test_vec.h"
 
+#ifndef PATH_MAX
```
p-steuer, line 33: why ifndef ?

src/fips.c
```diff
@@ -101,6 +121,201 @@ fips_init(void)
...
+		return NULL;
+
+	fn = strrchr(origpath, '/');
+	if (fn == NULL) {
+		fn = origpath;
+	} else {
+		++fn;
+	}
+
+	strncpy(path, origpath, fn - origpath);
```
p-steuer, line 168: why fn - origpath ?

src/fips.c
```diff
@@ -101,6 +121,201 @@ fips_init(void)
...
+	size_t hlen, len;
+	long keylen;
+
+	keybuf = OPENSSL_hexstr2buf(hmackey, &keylen);
+	pkey = EVP_PKEY_new_mac_key(EVP_PKEY_HMAC, NULL, keybuf, (int)keylen);
+	if (!pkey)
+		return rc;
+
+	mdctx = EVP_MD_CTX_create();
+	if (!mdctx)
```
p-steuer, line 193: i think pkey and keybuf not freed in this case

src/fips.c
```diff
@@ -101,6 +121,201 @@ fips_init(void)
...
+{
+	int rc;
+	char path[PATH_MAX];
+	const char *libname = "libica.so";
+	const char *symbolname = "ica_sha256";
+
+	rc = get_library_path(libname, symbolname, path, sizeof(path));
+	if (rc != 0) {
+		syslog(LOG_ERR, msg1);
+		fips |= ICA_FIPS_INTEGRITY;
```
p-steuer, line 309: you want to return in this case. otherwise the "cheks passed" masg will be written to the syslog.

src/fips.c
```diff
@@ -101,6 +121,201 @@ fips_init(void)
...
+	}
+
+	if (!FIPSCHECK_verify(path)) {
+		syslog(LOG_ERR, msg2, path);
+		fips |= ICA_FIPS_INTEGRITY;
```
p-steuer, line 314: same here

src/fips.c
jschmidb: Yes, in Libica it's currently the only occurrence of PATH_MAX. But as this constant is defined in linux limits.h, I thought better be safe ...

src/fips.c
jschmidb: For example, if origpath = /usr/local/lib/libica.so.3, then fn - origpath = 15, which is the length of the path without the file name. The resulting path will the be first /usr/local/lib/ and then be assembled to /usr/local/lib/.libica.so.3.hmac

src/fips.c
jschmidb: Done.

src/fips.c
jschmidb: Done.

src/fips.c
jschmidb: Done.

src/fips.c
p-steuer: okay

src/fips.c
p-steuer: okay

PR #51, opened 2020-08-27
FIPS: skip SHA3 tests if running on hardware without SHA3

jschmidb:

### DIFF
--- a/src/fips.c
+++ b/src/fips.c
@@ -95,11 +95,26 @@ SHA_KAT(384, 512);
 SHA_KAT(512, 512);
 #undef SHA_KAT
 
+static inline int sha3_available(void)
+{
+	sha3_224_context_t sha3_224_context;
+	unsigned char output_hash[SHA3_224_HASH_LENGTH];
+	unsigned char test_data[] = { 0x61,0x62,0x63 };
+	int rc = 0;
+
+	rc = ica_sha3_224(SHA_MSG_PART_ONLY, sizeof(test_data), test_data,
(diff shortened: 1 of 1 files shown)


p-steuer (approved):

ifranzki:
This is certainly a strait forward way to fix this issue, but I wonder if a better approach would be use ica_get_functionlist() and check the flags for the SHA3 mechanisms, if they are available in HW (or SW, but that's never the case as of today, because SHA3 does not have SW fallback). 

icainfo (which uses ica_get_functionlist()) shows that the flags are set correctly:

```
# icainfo
      Cryptographic algorithm support
-------------------------------------------
 function      |  hardware  |  software
---------------+------------+------------
...
      SHA3-224 |     no     |      no    <----  this is on a system that does not have SHA3 support in CPACF.
      SHA3-256 |     no     |      no
      SHA3-384 |     no     |      no
      SHA3-512 |     no     |      no
     SHAKE-128 |     no     |      no
     SHAKE-256 |     no     |      no
```
On a system thats has SHA3 support the HW column shows 'yes'.

jschmidb:
Good point Ingo ... but at least at the moment the function list gets initialized *after* running the power_up_tests, so we would have to re-order the functions. For now I would leave the fix as it is, also because SUSE and RH already confirmed that the fix works. But let's keep this in mind.

ifranzki:
> function list gets initialized after running the power_up_tests

good point. Yes leave it as is for now.

ifranzki (approved):

PR #53, opened 2020-09-01
FIPS: fix inconsistent error handling.

hramrach:
when the HMAC file is not present/readable it is also an error.

Signed-off-by: Michal Suchanek <msuchanek@suse.de>

### DIFF
--- a/src/fips.c
+++ b/src/fips.c
@@ -295,10 +295,8 @@ static int FIPSCHECK_verify(const char *path)
 		return 0;
 
 	fp = fopen(hmacpath, "r");
-	if (fp == NULL) {
-		rc = 1;
+	if (fp == NULL)
 		goto end;
-	}
 
 	if (getline(&known_hmac_str, &n, fp) <= 0)
 		goto end;


jschmidb:
Hmm yes, but when we treat this as an error, the hmac file would be required to be present in every case. What, if a distributor decides to not use the lib integrity check? My first approach implementing this integrity check was to make it optional. So when there is no .hmac file, the behavior is as before, but when it's there, the check must succeed.

hramrach:
Which is not the desired behavior. Maybe the fips functionality should be built-time option?
Will try to find out what other libraries with fpis support do.

hramrach:
https://gnutls.org/manual/gnutls.html#FIPS140_002d2-mode

jschmidb:
Looks like this fix would break all internal tests. We will discuss this in our team and get back to you then.

jschmidb:
Hi Michal, can you have a look at libica 3.8.0? I think your issue should be fixed there.

hramrach:
This has been implemented in 5dd1f6ad0ad5904223b9de5a2244bdbf4ba04e90

Thanks

PR #54, opened 2020-10-07
FIPS: enforce the hmac check

jschmidb:
When in FIPS mode, the .hmac file is now created during make and its existence
is enforced by changing the check in routine FIPSCHECK_verify to fail if it
cannot be opened.

This should make it easier for distributors to adapt the lib integrity check for their
distros, just by changing the static hmac key in fips.c and Makefile.am.

Notes:

- The openssl commandline expects a hex key
- The "check: all" in Makefile.am is necessary to create the .hmac file before running the internal tests when doing a "make check" without a "make all" before
- explicitly removing the .hmac file from src/.libs during "make uninstall" is not necessary, as the whole .libs directory is removed

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -4,6 +4,7 @@ SUBDIRS = doc include src test
 
 dist_doc_DATA = AUTHORS ChangeLog INSTALL LICENSE README.md
 EXTRA_DIST = libica.map libica.spec
+MAJOR := `echo $(VERSION) | cut -d. -f1`
 
 coverage: check
 	@echo -e "\n-----------------";
@@ -14,3 +15,14 @@ coverage: check
 	@echo -e "libica coverage";
 	@echo -e "---------------\n";
 	cd ${top_builddir}/src && gcov .libs/*.gcda
+
+if ICA_FIPS
+install-data-hook:
+	$(INSTALL) -m 0444 ${top_builddir}/src/.libs/.libica.so.$(VERSION).hmac $(DESTDIR)$(libdir)
+	cd $(DESTDIR)$(libdir) && ln -sf .libica.so.$(VERSION).hmac .libica.so.$(MAJOR).hmac
+
+uninstall-hook:
+	rm -f $(DESTDIR)$(libdir)/.libica.so.$(MAJOR).hmac
+	rm -f $(DESTDIR)$(libdir)/.libica.so.$(VERSION).hmac
+endif
+
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -1,6 +1,10 @@
 VERSION = 3:7:0
 
 AM_CFLAGS = @FLAGS@
+MAJOR := `echo $(VERSION) | cut -d: -f1`
+MINOR := `echo $(VERSION) | cut -d: -f2`
+PATCH := `echo $(VERSION) | cut -d: -f3`
+VERSION1 := $(MAJOR).$(MINOR).$(PATCH)
 
 # lib
 
@@ -29,6 +33,18 @@ libica_la_SOURCES = ica_api.c init.c icastats_shared.c s390_rsa.c \
 EXTRA_DIST = mp.pl
 mp.S	: mp.pl
 	./mp.pl mp.S
+
+if ICA_FIPS
+hmac-file-lnk: hmac-file
+	$(AM_V_GEN) cd ${top_builddir}/src/.libs && ln -sf .libica.so.$(VERSION1).hmac .libica.so.$(MAJOR).hmac
+
+hmac-file: libica.la
+	$(AM_V_GEN) openssl dgst -sha256 -mac hmac -macopt hexkey:00000000 ${top_builddir}/src/.libs/libica.so.$(VERSION1) | sed -e 's/^.* //' > ${top_builddir}/src/.libs/.libica.so.$(VERSION1).hmac
+
+hmac_files = hmac-file hmac-file-lnk
+
+all-local: $(hmac_files)
+endif
 
 # bin
 
@@ -69,3 +85,5 @@ internal_tests_ec_internal_test_SOURCES = \
 		    include/s390_rsa.h include/s390_sha.h include/test_vec.h \
 		    include/rng.h ../test/testcase.h
 endif
+
+.PHONY: hmac-file hmac-file-lnk
--- a/src/fips.c
+++ b/src/fips.c
@@ -296,7 +296,6 @@ static int FIPSCHECK_verify(const char *path)
 
 	fp = fopen(hmacpath, "r");
 	if (fp == NULL) {
-		rc = 1;
 		goto end;
 	}
 
--- a/src/s390_ecc.c
+++ b/src/s390_ecc.c
@@ -739,7 +739,7 @@ unsigned int ecdh_hw(ica_adapter_handle_t adapter_handle,
 		const ICA_EC_KEY *privkey_A, const ICA_EC_KEY *pubkey_B,
 		unsigned char *z)
 {
-	uint8_t *buf;
+	uint8_t *buf = NULL;
 	size_t len;
 	int rc;
 	struct ica_xcRB xcrb;
@@ -760,8 +760,10 @@ unsigned int ecdh_hw(ica_adapter_handle_t adapter_handle,
 		return EIO;
 
 	reply_p = make_ecdh_request(privkey_A, pubkey_B, &xcrb, &buf, &len);
-	if (!reply_p)
-		return EIO;
+	if (!reply_p) {
+		rc = EIO;
+		goto ret;
+	}
 
 	rc = ioctl(adapter_handle, ZSECSENDCPRB, xcrb);
 	if (rc != 0) {
@@ -777,8 +779,10 @@ unsigned int ecdh_hw(ica_adapter_handle_t adapter_handle,
 	memcpy(z, reply_p->raw_z_value, privlen);
 	rc = 0;
 ret:
-	OPENSSL_cleanse(buf, len);
-	free(buf);
+	if (buf) {
+		OPENSSL_cleanse(buf, len);
+		free(buf);
+	}
 	return rc;
 }
 
@@ -1125,7 +1129,7 @@ unsigned int ecdsa_sign_hw(ica_adapter_handle_t adapter_handle,
 		const ICA_EC_KEY *privkey, const unsigned char *hash, unsigned int hash_length,
 		unsigned char *signature)
 {
-	uint8_t *buf;
+	uint8_t *buf = NULL;
 	size_t len;
 	int rc;
 	struct ica_xcRB xcrb;
@@ -1153,8 +1157,10 @@ unsigned int ecdsa_sign_hw(ica_adapter_handle_t adapter_handle,
 
 	reply_p = make_ecdsa_sign_request((const ICA_EC_KEY*)privkey,
 			X, Y, hash, hash_length, &xcrb, &buf, &len);
-	if (!reply_p)
-		return EIO;
+	if (!reply_p) {
+		rc = EIO;
+		goto ret;
+	}
 
 	rc = ioctl(adapter_handle, ZSECSENDCPRB, xcrb);
 	if (rc != 0) {
@@ -1170,8 +1176,10 @@ unsigned int ecdsa_sign_hw(ica_adapter_handle_t adapter_handle,
 	memcpy(signature, reply_p->signature, reply_p->vud_len-8);
 	rc = 0;
 ret:
-	OPENSSL_cleanse(buf, len);
-	free(buf);
+	if (buf) {
+		OPENSSL_cleanse(buf, len);
+		free(buf);
+	}
 	return rc;
 }
 
@@ -1582,7 +1590,7 @@ unsigned int ecdsa_verify_hw(ica_adapter_handle_t adapter_handle,
 		const ICA_EC_KEY *pubkey, const unsigned char *hash, unsigned int hash_length,
 		const unsigned char *signature)
 {
-	uint8_t *buf;
+	uint8_t *buf = NULL;
 	size_t len;
 	int rc;
 	struct ica_xcRB xcrb;
@@ -1602,8 +1610,10 @@ unsigned int ecdsa_verify_hw(ica_adapter_handle_t adapter_handle,
 
 	reply_p = make_ecdsa_verify_request(pubkey, hash, hash_length,
 					    signature, &xcrb, &buf, &len);
-	if (!reply_p)
-		return EIO;
+	if (!reply_p) {
+		rc = EIO;
+		goto ret;
+	}
 
 	rc = ioctl(adapter_handle, ZSECSENDCPRB, xcrb);
 	if (rc != 0) {
@@ -1625,8 +1635,10 @@ unsigned int ecdsa_verify_hw(ica_adapter_handle_t adapter_handle,
 
 	rc = 0;
 ret:
-	OPENSSL_cleanse(buf, len);
-	free(buf);
+	if (buf) {
+		OPENSSL_cleanse(buf, len);
+		free(buf);
+	}
 	return rc;
 }
 
@@ -1940,7 +1952,7 @@ static int eckeygen_cpacf(ICA_EC_KEY *key)
  */
 unsigned int eckeygen_hw(ica_adapter_handle_t adapter_handle, ICA_EC_KEY *key)
 {
-	uint8_t *buf;
+	uint8_t *buf = NULL;
 	size_t len;
 	int rc;
 	struct ica_xcRB xcrb;
@@ -1959,8 +1971,10 @@ unsigned int eckeygen_hw(ica_adapter_handle_t adapter_handle, ICA_EC_KEY *key)
 		return ENODEV;
 
 	reply_p = make_eckeygen_request(key, &xcrb, &buf, &len);
-	if (!reply_p)
-		return EIO;
+	if (!reply_p) {
+		rc = EIO;
+		goto ret;
+	}
 
 	rc = ioctl(adapter_handle, ZSECSENDCPRB, xcrb);
 	if (rc != 0) {
@@ -1985,8 +1999,10 @@ unsigned int eckeygen_hw(ica_adapter_handle_t adapter_handle, ICA_EC_KEY *key)
 	memcpy(key->X, (char*)pub_p->pubkey, 2*privlen);
 	rc = 0;
 ret:
-	OPENSSL_cleanse(buf, len);
-	free(buf);
+	if (buf) {
+		OPENSSL_cleanse(buf, len);
+		free(buf);
+	}
 	return rc;
 }
 
--- a/src/s390_ecc.c
+++ b/src/s390_ecc.c
@@ -64,6 +64,18 @@ int scalar_mulx_cpacf(unsigned char *res_u,
 		      int curve_nid);
 
 /**
+ * Since kernel 4.10 the zcrypt device driver has multi domain support and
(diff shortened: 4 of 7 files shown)


ifranzki:
Are you aware that the travis build now fails:
https://travis-ci.com/github/opencryptoki/libica/builds/188536325

jschmidb:
Hi Ingo, yes I am, but I have not yet discovered why ....

jschmidb:
Guys, I just wanted your general opinion about this lib integrity check. As mentioned above, I don't know yet why the ec_internal_test fails. Here some more info about how this works:

- Up to now, it is no error if the .hmac file does not exist (see fips.c, line 299), but SUSE wants this be changed
- This pr now changes this, but as a consequence, the .hmac file must be created in any case, independent of a specific distro
- When doing a make check without a make (all) before, the internal tests failed, because the .hmac file wasn't created at this time.
- Therefore I added the "check: all" in Makefile.am to enforce make all, and thus creating the .hmac file, before actually running the internal tests.

Makefile.am
```diff
@@ -4,6 +4,7 @@ SUBDIRS = doc include src test
 
 dist_doc_DATA = AUTHORS ChangeLog INSTALL LICENSE README.md
 EXTRA_DIST = libica.map libica.spec
+MAJOR = `echo $(VERSION) | cut -d. -f1`
```
juergenchrist, line 7: Should be a `:=` here to not unnecessarily spawn a sub-shell whenever MAJOR is used.

Makefile.am
```diff
@@ -14,3 +15,18 @@ coverage: check
...
+
+if ICA_FIPS
+check: all
+	@echo make all and all-local before make check because of .hmac file
+
+all-local:
+	openssl dgst -sha256 -mac hmac -macopt hexkey:00000000 src/.libs/libica.so.$(VERSION) | sed -e 's/^.* //' > src/.libs/.libica.so.$(MAJOR).hmac
+
+install-data-hook:
+	cp src/.libs/.libica.so.$(MAJOR).hmac /usr/local/lib/
```
juergenchrist, line 27: Shouldn't we use libdir instead of hard-coding the install path?

Makefile.am
```diff
@@ -14,3 +15,18 @@ coverage: check
...
+
+uninstall-hook:
+	rm /usr/local/lib/.libica.so.$(MAJOR).hmac
```
juergenchrist, line 30: Use libdir?

ifranzki:
Travis fails due to
`FAIL: ../src/internal_tests/ec_internal_test`

When I run this test manually on my system, I get the same error with this output:
```
./ec_internal_test
ERROR: Hashing failed. (ECDSA test vector 0)
```

The code calls ica_sha1() and this returns with EACCES due to:
```
#ifdef ICA_FIPS
	if (fips >> 1)
		return EACCES;    
#endif /* ICA_FIPS */
```
Variable fips is 4, which is ICA_FIPS_INTEGRITY.
This is set in function fips_lib_integrity_check() because it fails to verify the hmac of /usr/lib64/.libica.so.3.7.0.
There is no file /usr/lib64/.libica.so.3.7.0.hmac, because I did not install the freshly built libica before running the tests.

The same happens in travis. Travis does not install it (would not be allowed to do so I guess), but runs the tests out of the source directory.

So it looks like there is some general problem with running the ICA tests without installing it first. One should really be able to run the tests out of the source directory without installing it first!

Maybe LD_LIBRARY_PATH must be set to point to src/.libs/ to use the just built library? 

But this does not work either:
It loads /root/libica-jschmidb/src/.libs/libica.so.3.7.0 in fips_lib_integrity_check(), and thus looks for file /root/libica-jschmidb/src/.libs/.libica.so.3.7.0.hmac for the HMAC.  But only /root/libica-jschmidb/src/.libs/.libica.so.3.hmac exists.....

src/fips.c
```diff
@@ -296,7 +296,7 @@ static int FIPSCHECK_verify(const char *path)
 
 	fp = fopen(hmacpath, "r");
 	if (fp == NULL) {
-		rc = 1;
+		rc = 0;
```
juergenchrist, line 299: You could remove this line since `rc` is still at the initial value of 0.

juergenchrist (commented):
Not sure about the Makefile.am changes.  I think we need to change the hard-coded paths.

src/fips.c
jschmidb: Yep, you are right ...

ifranzki:
There is one basic question to be answered: Is it required to do a 'make install' before you can do 'make check', so that it picks up the just built and just installed library, or is it supposed to be possible to run 'make check' without installing it? 

Dependent on the answer of this question, we need to either fix the code to allow 'make check' without installing (may require to manually set LD_LIBRARY_PATH though), or we need to fix the travis build to perform a 'make install' before running 'make check'.

jschmidb:
Yes I agree. It would be easier if "make check" would require a "make install" before. But in the past you could just do a "make check" against the local .so module in src/.libs. I still do not understand why the ec_internal_test wants to verify the hmac of /usr/lib64/.libica.so.3.7.0, but all other tests obviously not ...

Makefile.am
jschmidb: hmm yes, I agree, but $(LIBDIR) is not set here .... maybe we can pass it via libica.spec ...

jschmidb:
ok, when I do two things it works:

1. `# export LD_LIBRARY_PATH=/root/libica_fips/src/.libs`
2. provide a .libica.so.3.7.0.hmac in src/.libs in addition to the .libica.so.3.hmac

Makefile.am
jschmidb: ok, I found out that the variable is $(libdir) and not $(LIBDIR)! So I will change the hard coded /usr/local/lib to $(libdir).

p-steuer:
> There is one basic question to be answered: Is it required to do a 'make install' before you can do 'make check', so that it picks up the just built and just installed library, or is it supposed to be possible to run 'make check' without installing it?
> 
> Dependent on the answer of this question, we need to either fix the code to allow 'make check' without installing (may require to manually set LD_LIBRARY_PATH though), or we need to fix the travis build to perform a 'make install' before running 'make check'.

make check is intended to be run before installing. There is a make installcheck target for tests that only make sense to run against installed libs.

To make it work with make check, setting LD_LIBRARY_PATH should not be required, it is already set in the test environment (see test/Makefile.am).

p-steuer:
I just debugged this one. Basically, dladdr returns the name of the symlink (.3) in dli_fname, but the name of the shared object (.3.7.0) for the internal test. This may have to do with the internal test being special: all other tests are just executables that link libica. The internal test is the libica source build into an executable, that again links / dlopens libica, so the contructor runs multiple times ..

I suggest to provide an .libica.major.minor.patch.hmac file and a libica.major.hmac symlink to it, analoguous to the shared object.

Makefile.am
jschmidb: Hi Juergen, you mean 
MAJOR := `echo $(VERSION) | cut -d. -f1`
?

jschmidb:
ok, I'm now creating a .libica.so.3.7.0.hmac and a symlink .libica.so.3.hmac.
This causes the Travis checks to pass. I cannot reproduce the case where the LD_LIBRARY_PATH was necessary.

jschmidb:
One probably last question:
I have this target:
```
check: all
	@echo make all and all-local before make check because of .hmac file
```
and the @echo obviously does not do anything functional. But I saw that if I leave the @echo out, the target doesn't work. Any thoughts about this?

Makefile.am
juergenchrist: yes

jschmidb: ok, done.

Makefile.am
jschmidb: done.

src/fips.c
jschmidb: done.

p-steuer:
LD_LIBRARY_PATH has to be set if the tests executables are invoked by hand rather than by make check and the lib has not been installed.

> check: all

Shouldnt

> check-local: all-local

do the trick. I mean the -local targets are provided by autotools to add something to the standard targets like all, check etc, i would not mess with the latter directly if there is another option.

jschmidb:
No, unfortunately not. I think the reason is that I have to hook into make check after the modules have been built, but before the tests get executed. 
So "check" depends on "all" (which builds the modules), "all-local" comes after all and creates the .hmac file, and finally "check" executes the tests.

p-steuer:
But can you not just add the hmac file to libica_la_SOURCES and move what is done by all-local right now to a correspondig build target.

E.g. in src/makefile.am:
```
libica_la_SOURCES = ......  <hmacfilename>

<hmacfilename>:
        The commands from all-local go here
```

p-steuer:
No, you have to generate the file after the lib is build:

```
hmac_SOURCES = <hmacfile> libica.la
 
<hmacfile>:
     generate hmac file and symlink here

```

Makefile.am
```diff
@@ -14,3 +15,21 @@ coverage: check
 	@echo -e "libica coverage";
 	@echo -e "---------------\n";
 	cd ${top_builddir}/src && gcov .libs/*.gcda
+
+if ICA_FIPS
+check: all
+	@echo make all and all-local before make check because of .hmac file
+
+all-local:
```
juergenchrist, line 23: Shouldn't this be `all-hook`?  From what I know, -local targets are executed **before** the corresponding target and -hook targets are executed **after** the target.  In our case, we want the all target to build the library and then create the .hmac file.

juergenchrist: Also this rule should not be in this Makefile.am, but in src/Makefile.am as an all-hook.  Then I think it is executed at the correct place and you do not need the ``check: all`` artificial dependency.

Makefile.am
```diff
@@ -14,3 +15,21 @@ coverage: check
...
+	openssl dgst -sha256 -mac hmac -macopt hexkey:00000000 src/.libs/libica.so.$(VERSION) | sed -e 's/^.* //' > src/.libs/.libica.so.$(VERSION).hmac
+	cd ${top_builddir}/src/.libs && ln -sf .libica.so.$(VERSION).hmac .libica.so.$(MAJOR).hmac
+
+install-data-hook:
+	cp ${top_builddir}/src/.libs/.libica.so.$(VERSION).hmac $(libdir)
```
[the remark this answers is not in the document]
jschmidb: good point! We didn't consider permissions so far. It works with the given install cmdline and the .hmac file has the specified permissions.

Makefile.am
jschmidb: ok, I moved this target to src/Makefile.am:

- The `check: all` dependency is now in fact superfluous
- But the all-local cannot be replaced by all-hook

I realized that VERSION is defined in src/makefile.am as `VERSION = 3:7:0` and I wonder why colons are used here?
So I introduced a VERSION1 temporarily as `VERSION1=3.7.0`, but this is not what we want at the end.

src/Makefile.am
```diff
@@ -30,6 +32,12 @@ EXTRA_DIST = mp.pl
 mp.S	: mp.pl
 	./mp.pl mp.S
 
+if ICA_FIPS
+all-local:
```
juergenchrist, line 36: This is still very dangerous.  In a parallel build, there is no guarantee that libica.so is already build when this rule runs.  As a bare minimum, this rule should depend on $(LTLIBRARIES) to make sure, libtool has built the libraries.

jschmidb: Hm, yes, didn't think of parallel builds yet. I added the $(LTLIBRARIES) dependency and verified that e.g. a make -j8 succeeds.

jschmidb:
I changed src/Makefile.am again to derive major, minor, and patch level from the explicitly specified VERSION = 3:7:0. I'm then re-assembling the parts to VERSION1 = 3.7.0 (with dots). The additional variable VERSION1 is not really necessary, as we could also directly use the parts in the openssl and ln commandlines, but for me this is better readable.

Makefile.am
```diff
@@ -14,3 +15,14 @@ coverage: check
 	@echo -e "libica coverage";
 	@echo -e "---------------\n";
 	cd ${top_builddir}/src && gcov .libs/*.gcda
+
+if ICA_FIPS
+install-data-hook:
+	$(INSTALL) -m 0444 ${top_builddir}/src/.libs/.libica.so.$(VERSION).hmac $(libdir)
```
juergenchrist, line 21: Please make the build relocatable.  The install-folder should read `$(DESTDIR)$(libdir)`.  This is typically needed to build RPMs and otherwise has to be patched by distributors.

Makefile.am
```diff
@@ -14,3 +15,14 @@ coverage: check
...
+	cd $(libdir) && ln -sf .libica.so.$(VERSION).hmac .libica.so.$(MAJOR).hmac
```
juergenchrist, line 22: Same here: `cd $(DESTDIR)$(libdir)`.

jschmidb:
OK, also here I would like to continue our discussions and finalize this pull request. 
The last comment from Juergen asked for making the build relocatable by using `$(DESTDIR)$(libdir)`
I now added this change in Makefile.am for install-data-hook, but also for uninstall-hook.
Juergen, can you have a quick look?
I think there's nothing else left, if you (all) still see something, let me know.
I'm now pushing the changes in Makefile.am.

jschmidb:
Just added Harald for review: Harald can you have a look especially at commit "ECC: autoselect domain if supported by kernel"? This is the fix for an old bugzilla from Christian about ECC load balancing. The fix tries to use AUTOSELECT DOM if the kernel supports it. I tested this - of course with a new kernel - but have no system with a kernel not supporting AUTOSELECT DOM.

src/Makefile.am
```diff
@@ -30,6 +34,12 @@ EXTRA_DIST = mp.pl
 mp.S	: mp.pl
 	./mp.pl mp.S
 
+if ICA_FIPS
+all-local: $(LTLIBRARIES)
```
juergenchrist, line 38: Shouldn't this be something like
```
.libs/.libica.so.$(VERSION1).hmac: $(LTLIBRARIES):
	$(AM_V_GEN) openssl dgst -sha256 ...

.libs/.libica.so.$(MAJOR).hmac: .libs/.libica.so.$(VERSION1).hmac
	cd .libs && ln -sf .libica.so.$(VERSION1).hmac .libica.so.$(MAJOR).hmac

all-local: .libs/libica.so.$(VERSION1).hmac .libs/.libica.so.$(MAJOR).hmac
```

jschmidb:
Update to commit "ECC: autoselect domain if supported by kernel": Harald provided me an old RHEL 7.4 with a kernel that does not support autoselect domain. Thanks Harald! I checked the code change here and it looks good. The code first tries to use autoselect, gets an rc = -1 from ioctl, then determines the default domain, re-tries ioctl, and ioctl succeeds as expected. On kernels with autoselect domain support, the first try using autoselect succeeds.

src/Makefile.am
jschmidb: Following up our internal discussion I'm pushing some changes. Coding the two hmac file targets like above causes build errors, so I'm just using:
```
hmac-file-lnk: hmac-file
	$(AM_V_GEN) cd ${top_builddir}/src/.libs && ln -sf .libica.so.$(VERSION1).hmac .libica.so.$(MAJOR).hmac

hmac-file: libica.la
	$(AM_V_GEN) openssl dgst -sha256 -mac hmac -macopt hexkey:00000000 ${top_builddir}/src/.libs/libica.so.$(VERSION1) | sed -e 's/^.* //' > ${top_builddir}/src/.libs/.libica.so.$(VERSION1).hmac

hmac_files = hmac-file hmac-file-lnk

all-local: $(hmac_files)  
```

juergenchrist: But then `hmac-file-lnk` and `hmac-file` should be PHONY-targets since they do not actually create files with exactly that name.

jschmidb: OK, added .PHONY for the two targets

src/Makefile.am
```diff
@@ -69,3 +85,5 @@ internal_tests_ec_internal_test_SOURCES = \
 		    include/s390_rsa.h include/s390_sha.h include/test_vec.h \
 		    include/rng.h ../test/testcase.h
 endif
+
+.PHONY: hmac-file hmac-file-lnk
```
juergenchrist, line 89: Missing newline at end of file

src/s390_ecc.c
```diff
@@ -63,6 +63,19 @@ int scalar_mulx_cpacf(unsigned char *res_u,
...
+ * accepts CPRBs via the ioctl ZSECSENDCPRB with domain addressing 0xFFFF
+ * (AUTOSELECT_DOM in zcrypyt.h). This allows for load balancing between
+ * multiple available crypto cards.
+ */
+typedef enum {
+	dom_addressing_not_set = 0,
+	dom_addressing_autoselect,
+	dom_addressing_default_domain,
+} dom_addressing_t;
+int dom_addressing = dom_addressing_not_set;
```
juergenchrist, line 77: Why do we use `dom_addressing_not_set` as default here?  If I understand the remaining code correctly, on all paths where we evaluate `dom_addressing`, we first check if it is equal to `dom_addressing_not_set` and, if so, change it to `dom_addressing_autoselect`.  Couldn't we simply initialize the variable already to `dom_addressing_autoselect` and get rid of the `if` statements to change the value if not already set?  Then, we could also get rid of `dom_addressing_not_set` since it is not used anywhere anymore.  The sequence seems to be "if nothing set, set and try autoselect, if that fails set and try default domain".

juergenchrist (approved):

src/s390_ecc.c
jschmidb: Yes, indeed. This saves some code. I made the change and re-tested on both new and old kernel.

src/Makefile.am
jschmidb: Newline added. Didn't know there is one required.

ifranzki (approved):

p-steuer (approved):

src/s390_ecc.c
```diff
@@ -63,6 +63,18 @@ int scalar_mulx_cpacf(unsigned char *res_u,
...
+ * accepts CPRBs via the ioctl ZSECSENDCPRB with domain addressing 0xFFFF
+ * (AUTOSELECT_DOM in zcrypyt.h). This allows for load balancing between
+ * multiple available crypto cards.
+ */
+typedef enum {
+	dom_addressing_autoselect = 0,
+	dom_addressing_default_domain,
+} dom_addressing_t;
+int dom_addressing = dom_addressing_autoselect;
+
```
hfreude, line 77: static ?

src/s390_ecc.c
```diff
@@ -767,8 +781,18 @@ unsigned int ecdh_hw(ica_adapter_handle_t adapter_handle,
...
+		if (!reply_p) {
+			rc = EIO;
+			goto ret;
+		}
+
+		rc = ioctl(adapter_handle, ZSECSENDCPRB, xcrb);
+		if (rc != 0) {
+			rc = EIO;
+			goto ret;
+		}
```
hfreude, line 795: this duplicates the code ... and uses dom_addressing as a side effect.
what about 

typedef enum {
	dom_addressing_autoselect = 0,
	dom_addressing_default_domain = 1,
} dom_addressing_t;

...

for (domaddr = dom_addressing_autoselect; domaddr <= dom_addressing_default_domain; domaddr++) {
  ...
}
if (domaddr >dom_addressing_default_domain) {
  // both did not work ...
}

still with side-effect but the code is just once.

Btw. what's with reply_p pointing to ? Your code gives me the impression, it is not freed when the first attempt fails.

hfreude (commented):
as wrote - consider changing the retries to some kind of a loop to avoid doubling the code.

PR #55, opened 2020-10-29
Fips ctr check

jschmidb:


src/Makefile.am
```diff
@@ -9,7 +9,7 @@ lib_LTLIBRARIES = libica.la
 libica_la_CFLAGS = ${AM_CFLAGS} -I${srcdir}/include -I${srcdir}/../include \
 		   -fvisibility=hidden -pthread
 libica_la_CCASFLAGS = ${AM_CFLAGS}
-libica_la_LIBADD = @LIBS@ -lrt -lcrypto -ldl
+libica_la_LIBADD = @LIBS@ -lrt -lcrypto -ldl -lm
```
p-steuer, line 12: Why do we link libmath ? As far as i can tell you only need it for "pow(2, x)" which you could easily do just using the shift operator "2 << x" ...

p-steuer (changes requested):

src/ica_api.c
```diff
@@ -2272,6 +2273,9 @@ unsigned int ica_des_ctr(const unsigned char *in_data, unsigned char *out_data,
 #ifdef ICA_FIPS
 	if (fips)
 		return EACCES;
+	if (ctr_width < (DES_BLOCK_SIZE*8) &&
```
ifranzki, line 2276: Blanks between DES_BLOCK_SIZE*8 -> DES_BLOCK_SIZE * 8
Not sure if the brackets are really needed.
Same at the other 2 changed places

src/Makefile.am
jschmidb: Hm .. I like the idea and did some tests. To calculate pow(2,n) we would have to code (1ULL << n).
```
printf("  1 << 31 = %llu\n",(1ULL << 31));
printf("  1 << 32 = %llu\n",(1ULL << 32));
printf("  1 << 63 = %llu\n",(1ULL << 63));
printf("  1 << 64 = %llu\n",(1ULL << 64));
printf("  pow(2,64) = %f\n",pow(2,64));
```
The output is:
```
  1 << 31 = 2147483648
  1 << 32 = 4294967296
  1 << 63 = 9223372036854775808
  1 << 64 = 0
  pow(2,64) = 18446744073709551616.000000
```
So we have a problem if the ctr_width is 64. And with an AES counter the numbers can get even bigger as the AES block size is 16. What do you think?

src/ica_api.c
jschmidb: OK, will add the blanks. I copied the code piece from a few lines below where they coded it like this. Maybe we can do some general cleanup later with a separate commit. I would leave the brackets for better readability.

src/Makefile.am
p-steuer: So IMO the check
```
data_length > pow(2,ctr_width)
```
is incorrect anyways: data_length are bytes not blocks, so data_length should be divided by block size (and ceil-ed if it is not a multiple of block size), since the counter is incremented per processed block:
```
(data_length + BLOCK_SIZE - 1) / BLOCK_SIZE > ...
```

Yes, the shift operator is limited, but the left hand side of above comparison can never exceed ULL max.
I think you should rethink the check. 

Also, the the counter input ctr may not be zero. If the function can be called in a streaming fashion, you have to consider that also.

src/ica_api.c
```diff
@@ -2761,6 +2767,9 @@ unsigned int ica_aes_ctr(const unsigned char *in_data, unsigned char *out_data,
 #ifdef ICA_FIPS
 	if (fips >> 1)
```
juergenchrist, line 2768: Here we right-shift by one.  In line 2273 we use `fips` directly.  Is this desired?

src/ica_api.c
```diff
@@ -2761,6 +2767,9 @@ unsigned int ica_aes_ctr(const unsigned char *in_data, unsigned char *out_data,
...
 		return EACCES;
+	if (ctr_width < (AES_BLOCK_SIZE*8) &&
+		data_length / AES_BLOCK_SIZE > (1ULL << ctr_width))
```
juergenchrist, line 2771: This does not work!  `AES_BLOCK_SIZE` is 16.  So we check that we shift less than 16*8 bits, but we only have a 64bit value.  So every shift >= 64 does not make sense.  Since `data_length / AES_BLOCK_SIZE` is strictly less than 2^64, why don't you replace the check with a hard comparison against 64 and then the shift?  Also, what about padding?  If `data_length` is not a multiple of `AES_BLOCK_SIZE`, please use this check:
```
if (ctr_width < 64U && ((data_length / AES_BLOCK_SIZE) + !!(data_length % AES_BLOCK_SIZE)) > (1UL << ctr_width))
```
This makes sure that `ctr_width` greater equal 64 bits is not a problem (since we cannot have that many blocks), and that the number of blocks needed (including possibly padded blocks via the modulo and double negation) still does not cause repeated counter values.

Similar check could be used for DES above.

juergenchrist (changes requested):

include/ica_api.h
```diff
@@ -2319,9 +2322,12 @@ unsigned int ica_3des_cfb(const unsigned char *in_data, unsigned char *out_data,
  * key if data_length used in the preceding call is a multiple of the cipher
  * block size.
  * @param ctr_width
- * A number U between 1 and cipher block size. The value is used by the counter
- * increment function which increments a counter value by incrementing without
- * carry the least significant U bytes of the counter value.
+ * A number U between 8 and cipher block size in bits. The value is used by the
+ * counter increment function which increments a counter value by incrementing
+ * without carry the least significant U bytes of the counter value. The value
```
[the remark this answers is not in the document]
jschmidb: Good catch! Changed.

src/ica_api.c
```diff
@@ -2272,6 +2273,9 @@ unsigned int ica_des_ctr(const unsigned char *in_data, unsigned char *out_data,
 #ifdef ICA_FIPS
 	if (fips)
 		return EACCES;
+	if (ctr_width < (DES_BLOCK_SIZE*8) &&
+		data_length > pow(2,ctr_width))
```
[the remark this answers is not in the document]
jschmidb: Yes, pow(2,n) is now replaced by 1UL << n.

src/ica_api.c
jschmidb: Yes, DES is disabled in FIPS mode, but 3DES not. So we can simply remove the check from the ica_des_ctr routine.

src/ica_api.c
jschmidb: Yes this makes sense:
```
if (ctr_width < (64*8) &&
		data_length / 64 > (1ULL << ctr_width))
```
In fact, using the DES_BLOCK_SIZE constant in the ica_3des_ctr routine (btw I removed the check completely for ica_des_ctr, because DES is blocked in fips mode anyway) is not correct. By coincidence, DES_BLOCK_SIZE is 64 bits, but it means in fact sizeof(unsigned long).

juergenchrist: I don't understand why you need `sizeof(unsigned long)` here?  Isn't the FIPS requirement for CTR-mode that the counter is not allowed to have the same value for different blocks?  So we need to divide the (possibly padded) data length by the blocksize and check if there are enough counter bits to represent that number.  That was what I tried to represent in the formula above.  

Of course, this only works if all data is passed in one go to this function.  If the function can be called multiple times in a streaming fashion, then we still need another additional check.  I am not sure about the usage pattern of this function, but I assume it is only used to de- or encrypt a whole message at once.

jschmidb:
After quite some time I now would like to get back to this pull request. OK, where are we and what is still to be done here?
Juergen commented on the ctr check and he is right: datalen is allowed to be not a multiple of the cipher blocksize and the check therefore must be:
```
if (ctr_width < 64U &&
    (data_length / blocksize + data_length % blocksize) > (1ULL << ctr_width))
    return EINVAL;
```
This applies for 3DES and AES. The check is not necessary for DES, as DES is blocked
in FIPS mode anyway. 
I'm now pushing the two changes in ica_api.c, Juergen can you have a quick look?
I think this was the last leftover for this pr.

juergenchrist:
> if (ctr_width < 64U &&
>     (data_length / blocksize + data_length % blocksize) > (1ULL << ctr_width))
>     return EINVAL;

I don't understand this.  You do not need one counter value for every byte in the non-blocksize block, but only one additional counter value if there is a block at the end that needs to get padded.  That is why I had the double negation in my formula above.

juergenchrist:
General problem:  Is this allowed:
```
ica_3des_ctr(in_part1, out_part1, part1_length, key,
                        // Use same counter settings...
			 ctr, ctr_width,
			 1);
ica_3des_ctr(in_part2, out_part2, part2_length, key,
                        // ... in multiple operations
			 ctr, ctr_width,
			 1);
```
If yes, how does your FIPS-Check deal with this?  Now I could have the same counter value for different blocks in successive calls.  I could not find anything about this function being a one-shot and not a streaming function.

jschmidb:
Good point! As discussed via phone:
- The ctr check is now reworked to correctly handle full data blocks and additional bytes in a last incomplete block.
- Chained calls are allowed, this is documented in the Libica book. I have no idea how we could detect counter wraps over a series of chained calls and would document this fact in the User's Guide. We can only guarantee that the ctr doesn't wrap during the execution of one single function. Patrick, what do you think here?

jschmidb:
fyi: I added an obvious bugfix from Christian Borntraeger. Didn't want to make a separate pr for this.

juergenchrist (approved):

jschmidb:
I think this one should be ok now. Just give me your final sign-offs if there's nothing else to discuss.

ifranzki (approved):

src/ica_api.c
```diff
@@ -2759,8 +2767,16 @@ unsigned int ica_aes_ctr(const unsigned char *in_data, unsigned char *out_data,
 	unsigned int function_code;
 
 #ifdef ICA_FIPS
+	/* Note that the FIPS ctr check cannot detect ctr wraps
+	 * over chained calls to this function. */
+	unsigned long num_blocks = data_length / DES_BLOCK_SIZE;
```
p-steuer, line 2772: DES_BLOCK_SIZE ?

src/ica_api.c
```diff
@@ -2759,8 +2767,16 @@ unsigned int ica_aes_ctr(const unsigned char *in_data, unsigned char *out_data,
...
+	unsigned int num_additional_bytes = data_length % DES_BLOCK_SIZE;
```
p-steuer, line 2773: DES_BLOCK_SIZE ?

src/ica_api.c
```diff
@@ -2504,10 +2504,18 @@ unsigned int ica_3des_ctr(const unsigned char *in_data, unsigned char *out_data,
 			 unsigned int direction)
 {
 #ifdef ICA_FIPS
+	/* Note that the FIPS ctr check cannot detect ctr wraps
```
p-steuer, line 2507: why is the check not added to ica_des_ctr altough the commit msg says so ?

src/ica_api.c
```diff
@@ -2759,8 +2767,16 @@ unsigned int ica_aes_ctr(const unsigned char *in_data, unsigned char *out_data,
...
+	 * over chained calls to this function. */
+	unsigned long num_blocks = data_length / DES_BLOCK_SIZE;
+	unsigned int num_additional_bytes = data_length % DES_BLOCK_SIZE;
 	if (fips >> 1)
 		return EACCES;
+	if (num_additional_bytes > 0)
+		num_blocks++;
+	if (ctr_width < 64U && num_blocks > (1ULL << ctr_width))
```
p-steuer, line 2778: num_blocks >= (1ULL << ctr_width) ?

src/ica_api.c
```diff
@@ -2504,10 +2504,18 @@ unsigned int ica_3des_ctr(const unsigned char *in_data, unsigned char *out_data,
...
+	 * over chained calls to this function. */
+	unsigned long num_blocks = data_length / DES_BLOCK_SIZE;
+	unsigned int num_additional_bytes = data_length % DES_BLOCK_SIZE;
 	if (fips >> 1)
 		return EACCES;
 	if (fips_check_3des_key((ica_des_key_triple_t *)key))
 		return EINVAL;
+	if (num_additional_bytes > 0)
+		num_blocks++;
+	if (ctr_width < 64U && num_blocks > (1ULL << ctr_width))
```
p-steuer, line 2517: num_blocks >= (1ULL << ctr_width) ?

src/ica_api.c
jschmidb: Because DES is blocked in FIPS mode anyway. But good catch on the commit msg! I will change it.

src/ica_api.c
jschmidb: num_blocks == 2^ctr_width works. I just tested it again.

src/ica_api.c
jschmidb: see above.

src/ica_api.c
p-steuer: > DES is blocked in FIPS mode anyway

Yeah, right didnt think of that. So its just the commit msg

src/ica_api.c
p-steuer: ah yeah i see. "off by one" in my head

src/ica_api.c
jschmidb: Good catch! Fixed.

p-steuer (approved):

PR #56, opened 2021-05-04
FIPS BUILD: check for openssl app

juergenchrist:
The openssl binary is needed for FIPS build to compute the HMAC.  Since the
computation of the HMAC calls openssl inside a piped command, a missing
openssl binary will not even produce a build error.  Instead, the HMAC file
will be empty and the library will not load in FIPS mode.

Instead of build error, lets make this a configure error and check during
configuration that the binary "openssl" is available.

Signed-off-by: Juergen Christ <jchrist@linux.ibm.com>


jschmidb:
Perfect, this was missing.

PR #57, opened 2021-05-05
use C99 macro for 64-bit integral type

sharkcz:
The U64() macro has been removed from OpenSSL 3.0, use a C99 standard
macro instead.

This fixes one part of https://bugzilla.redhat.com/show_bug.cgi?id=1952946


p-steuer:
Ah yeah, i did a similar change to openCryptoki a while ago. We should also build libica with openssl master to make sure it works with 3.0.

p-steuer (approved):

jschmidb (approved):
lgtm

juergenchrist (approved):

PR #59, opened 2021-05-10
New options

jschmidb:
This pull request introduces two new compile options NO_SW_FALLBACKS and NO_CPACF. They are used to build an additional libica module libica-cex, which only provides hw crypto based on crypto cards.


jschmidb:
There are Travis build errors ... will check.

ifranzki:
Commit "FIPS BUILD: check for openssl app" is already on master, you should probably do a rebase of your branch to only include those commits that are not already on master.

jschmidb:
I added two patches from Juergen: provide a separate icainfo-cex and find libica from phdrs. There was a minor change necessary to fix the icainfo-cex output: compile icainfo-cex with -DNO_SW_FALLBACKS and -DNO_CPACF.

jschmidb:
Travis build also fixed: the reason was a missing -DLIBNAME=\"libica\" for the internal tests.

include/ica_api.h
```diff
@@ -37,7 +37,7 @@
 #define ICA_EXPORT __attribute__((__visibility__("default")))
 #define ICA_DEPRECATED __attribute__((deprecated))
 
-#if defined(NO_SW_FALLBACKS)
+#if defined(NO_SW_FALLBACKS) || defined(NO_CPACF)
 #define UNUSED(var)			((void)(var))
```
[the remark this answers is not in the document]
ifranzki: BTW: this define is already in the 1st commit, so please remove/fix this in both commits.

ifranzki:
Travis failed again on the latest update.

jschmidb:
Had to revert a local testfix ...

include/ica_api.h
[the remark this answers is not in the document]
jschmidb: Yep, after all it were 3 places: in ica_api.c, s390_rsa.c, and testcase.h for all tests. 
Pushing again ...

jschmidb:
I added a commit to provide a separate manpage for icainfo-cex. Also changed the libica / libica-cex copyright year from 2016 to 2021.

src/fips.c
```diff
@@ -343,7 +392,7 @@ static void fips_lib_integrity_check(void)
 {
 	int rc;
 	char path[PATH_MAX];
-	const char *libname = "libica.so." VERSION;
+	const char *libname = LIBNAME ".so." VERSION;
```
juergenchrist, line 395: For a library that is `dlopen`ed, e.g., in ibmca engine, I find this check highly dangerous.  Assume you `dlopen` a libica you built yourself under `/home/myuser/libica.so` with the full path.  Then, this check will lead to an integrity check on the system installed libica which is not at all what will be used by the application.  I would instead suggest to `dl_iterate_phdr` over all PHDRs to search for the loaded library.  Of course, the `internal_test` is a major problem since it wants the fips-check to pass but does not even use the library and instead compiles all the files again.  But in this case, I guess the local compilation could also just add a preprocessor flag to bypass the fips check successfully.  Makes the code uglier, but should be the better solution.

jschmidb: Just pushed your [patch 2/2] implementing the `dl_iterate_phdr` as discussed, thanks! We must not run the library integrity check for internal tests. 
I also adjusted the icainfo-cex output again to reflect that disabling some functionality via NO_CPACF or NO_SW_FALLBACKS outweights blocking an algorithm via fips build. So when in FIPS mode, icainfo-cex displays '-' rather than 'blocked'.

jschmidb:
Pushed a small correction of the icainfo-cex manpage, showing the sample output: display '-' instead of 'blocked' according to the recent change in icainfo.c.

src/fips.c
```diff
@@ -159,6 +209,7 @@ fips_init(void)
 		FIPS_mode_set(1);
 	}
 }
+#if 0
```
juergenchrist, line 212: I guess this is my bad.  We should probably remove this code and not just comment it out.

src/ica_api.c
```diff
@@ -3312,16 +4128,22 @@ int ica_aes_gcm_kma_verify_tag(const unsigned char* known_tag, unsigned int tag_
 		return EFAULT;
 
 	return 0;
+#endif /* NO_CPACF */
 }
 
 void ica_aes_gcm_kma_ctx_free(kma_ctx* ctx)
 {
+#ifdef NO_CPACF
```
juergenchrist, line 4136: Do we really need this here?  The function is safe for calls with `NULL`.  So I think we could just use this function always in the same way without the `#ifdef NO_CPACF`.

juergenchrist (approved):

src/fips.c
jschmidb: Yes, makes sense. Removed.

src/ica_api.c
jschmidb: Good catch! It's the only function that can indeed be used without a change in libica-cex. 
Done.

ifranzki (approved):

PR #60, opened 2021-05-24
fix handling of --disable-foo options

sharkcz:
Currently passing eg. --disable-debug actually enables the debug build
as one would expect from --enable-debug. The fix is to omit setting the
enable_foo variable as the "action-if-given" parameter of AC_ARG_ENABLE,
because it handles both the --enable and --disable forms.

Signed-off-by: Dan Horák <dan@danny.cz>


jschmidb:
Good catch! Thanks Dan.

PR #63, opened 2021-06-08
Misc ECC fixes

jschmidb:
This pr provides several fixes for EC support in libica and addresses opencryptoki issue #404: Libica must check at key generate/init/derive if the EC curve is supported. It also adds openssl based EC key checking to prevent against initializing with invalid/corrupted EC key parts. Libica EC tests are adapted to the newly introduced rcs to skip tests if curves are not supported.

### DIFF
--- a/src/ica_api.c
+++ b/src/ica_api.c
@@ -1338,6 +1338,8 @@ int ica_ec_key_init(const unsigned char *X, const unsigned char *Y,
 #ifdef ICA_FIPS
 	if (fips >> 1)
 		return EACCES;
+	if (!curve_supported_via_openssl(key->nid))
+		return EPERM;
 #endif /* ICA_FIPS */
 
 	/* check for obvious errors in parms */
@@ -1378,6 +1380,8 @@ int ica_ec_key_generate(ica_adapter_handle_t adapter_handle, ICA_EC_KEY *key)
 #ifdef ICA_FIPS
 	if (fips >> 1)
 		return EACCES;
+	if (!curve_supported_via_openssl(key->nid))
+		return EPERM;
 #endif /* ICA_FIPS */
 
 	/* check for obvious errors in parms */
@@ -1430,6 +1434,8 @@ int ica_ecdh_derive_secret(ica_adapter_handle_t adapter_handle,
 #ifdef ICA_FIPS
 	if (fips >> 1)
 		return EACCES;
+	if (privkey_A != NULL && !curve_supported_via_openssl(privkey_A->nid))
+		return EPERM;
 #endif /* ICA_FIPS */
 
 	/* check for obvious errors in parms */
--- a/src/ica_api.c
+++ b/src/ica_api.c
@@ -1312,10 +1312,6 @@ ICA_EC_KEY* ica_ec_key_new(unsigned int nid, unsigned int *privlen)
 		return NULL;
 #endif /* ICA_FIPS */
 
-	/* check for obvious errors in parms */
-	if (!curve_supported(nid))
-		return NULL;
-
 	if ((key = malloc(sizeof(ICA_EC_KEY))) == NULL)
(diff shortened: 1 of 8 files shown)


src/s390_ecc.c
```diff
@@ -859,6 +867,9 @@ unsigned int ecdh_sw(const ICA_EC_KEY *privkey_A, const ICA_EC_KEY *pubkey_B,
 	if (rc == 0)
 		goto err;
 
+	if (!EC_KEY_check_key(a) || !EC_KEY_check_key(b))
```
ifranzki, line 871: I think this is superfluous here. You get ta and b from make_eckey() and make_public_eckey() above, and both do the check already. So no need to check them again here.

src/s390_ecc.c
```diff
@@ -2108,6 +2119,9 @@ unsigned int eckeygen_sw(ICA_EC_KEY *key)
 	if (!EC_KEY_generate_key(a))
 		goto err;
 
+	if (!EC_KEY_check_key(a))
```
ifranzki, line 2123: You don't need to check the key right after it was generated.

src/s390_ecc.c
jschmidb: Yes that's true. Removed.

src/s390_ecc.c
jschmidb: OK, removed.

test/testcase.h
```diff
@@ -190,6 +196,47 @@ is_supported_openssl_curve(int nid)
...
+
+static inline int sw_fallbacks_implemented(int nid)
+{
+	switch (nid) {
+	case NID_X9_62_prime192v1:
+	case NID_secp224r1:
+	case NID_X9_62_prime256v1:
+	case NID_secp384r1:
+	case NID_secp521r1:
+	case NID_brainpoolP160r1:
```
juergenchrist, line 208: Above you have the brainpool curves guarded by `#if OPENSSL_VERSION_NUMBER >= 0x010002000`.  Shouldn't we have the same guard here?

src/ica_api.c
```diff
@@ -1337,20 +1333,29 @@ ICA_EC_KEY* ica_ec_key_new(unsigned int nid, unsigned int *privlen)
...
+		return EPERM;
 #endif /* ICA_FIPS */
 
 	/* check for obvious errors in parms */
 	if (key == NULL)
 		return EINVAL;
 
+	/* check if curve is supported by hw */
+	if (!(curve_supported_via_online_card(key->nid) ||
+		  curve_supported_via_cpacf(key->nid)))
```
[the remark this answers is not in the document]
jschmidb: Fixed in routine curve_supported_via_cpacf. Thanks!

src/ica_api.c
```diff
@@ -1411,18 +1428,29 @@ int ica_ecdh_derive_secret(ica_adapter_handle_t adapter_handle,
...
+	if (privkey_A == NULL || pubkey_B == NULL)
+		return EINVAL;
+
+	privlen = privlen_from_nid(privkey_A->nid);
 	if (z == NULL || z_length < privlen || privkey_A->nid != pubkey_B->nid)
 		return EINVAL;
 
+	/* check if curve is supported by hw */
+	if (!(curve_supported_via_online_card(privkey_A->nid) ||
+		  curve_supported_via_cpacf(privkey_A->nid)))
```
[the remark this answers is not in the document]
jschmidb: Fixed, see above.

test/testcase.h
jschmidb: Yes, that's true, fixed.

juergenchrist (approved):

ifranzki (approved):
I looked mainly at commit "Add openssl key checks" - approved.

PR #64, opened 2021-06-09
FIPS: use fipshmac when available.

hramrach:
When patched with the distribution key the openssl script creating hmac with zero key does not work and all tests fail.

This works for me but not sure if it's the best way to fix.

I considered writing a test program that includes fips.c and prints the key but that does not work with cross-compilation.

Storing the key in a separate file so that it can be both included in fips.c and used in the openssl script might be an option.


jschmidb:
> Storing the key in a separate file so that it can be both included in fips.c and used in the openssl script might be an option.

Yes indeed. At the moment the hmac key must be specified/changed at two places. We may move the key into a separate file, as you suggest. I'm not sure about introducing a new dependency to fipshmac, but I'll check.

hramrach:
The idea of this patch is that on fips-free distribution or in CI environment you can test with openssl and when you use fips you will likely have fipshmac available. Not a great solution only checking the existence of the tool, though. It seems other utilities for fips handling exist, too.

hramrach:
This passes both on the CI and in the distro with FIPSHMAC=fipshmac and looks like overall improvement to readability.

PR #65, opened 2021-06-15
[TEST] Use build-tree icastats in icastats_test

juergenchrist:
icastats_test used a system installed icastats instead of the build-tree
icastats.  So we actually did not test the icastats binary but still tested
the correct counting.  Use the icastats binary from the build-tree to also
test that binary.

Signed-off-by: Juergen Christ <jchrist@linux.ibm.com>


jschmidb (approved):
I applied and tested the patch on my test system, all looks good, thanks!

PR #68, opened 2021-07-29
TESTCASES: add test for libica-cex

jschmidb:
This test checks the functionlist of the libica-cex module. All CPACF-related functions and software fallbacks are disabled here, so the mech mode flags must all be zero for these.


juergenchrist (approved):
lgtm

jschmidb:
Thanks!

Issue #61, opened 2021-05-24
eddsa test fails when built with OpenSSL 3.0


sharkcz:
The eddsa test fails when libica is built with OpenSSL 3.0. There is no problem when OpenSSL 1.1.1j is used. The environment is CentOS 9 Stream (aka to-be-RHEL-9) on a z15 machine, both LPAR and a KVM guest. My test build is https://kojihub.stream.centos.org/koji/taskinfo?taskID=272190

`test/eddsa_test.log` shows
````
ed25519_pc failed (eddsa_test.c:328): EVP_DigestVerify failed.
FAIL eddsa_test (exit status: 1)
````

jschmidb:
Same as for issue #58. Let's keep this on our list. Sooner or later we will provide updates for OpenSSL 3.0. for all of our libs.

jschmidb:
Hi Dan, today I took some time to look into this and saw that EVP_DigestVerify was used in eddsa_test without a preceding EVP_DigestVerifyInit. There are two places: one for ED25519 and one for ED448. 
I tested with a RHEL 9.0 from July 2021. On my test system all tests now run ok. 
Can you have a quick look at the commit? See https://github.com/opencryptoki/libica/pull/69

jschmidb:
Fixed by https://github.com/opencryptoki/libica/pull/69
Closing.

PR #69, opened 2021-08-04
TESTCASES: EVP_DigestVerifyInit was not called in eddsa_test

jschmidb:
EVP_DigestVerify was called without a EVP_DigestVerifyInit.
This bug shows up when using openssl 3.0.

This fix solves issue https://github.com/opencryptoki/libica/issues/61


sharkcz (approved):
Yup, it was the missing piece.

all tests passing in https://kojihub.stream.centos.org/koji/taskinfo?taskID=528381

jschmidb:
Thanks Dan!

PR #71, opened 2021-08-09
libica: fix wrong padding in rsa_crt_check function

hfreude:
When the qinv value has a leading 0, the padding
of this value in the ica_rsa_key_crt_t struct was
not currectly done in the ica_rsa_crt_key_check()
function. Fixed and tested, and works fine now.

Please note, for newer Openssl there is a simpler way to have the right padding:

BN_bn2binpad(bn_invq, tmp_buf, rsa_key->key_length/2);h

### DIFF
--- a/src/ica_api.c
+++ b/src/ica_api.c
@@ -1224,8 +1224,11 @@ unsigned int ica_rsa_crt_key_check(ica_rsa_key_crt_t *rsa_key)
 		/* qInv = (1/q) mod p */
 		BN_mod_inverse(bn_invq, bn_q, bn_p, ctx);
 		memset(tmp_buf, 0, rsa_key->key_length/2);
-		BN_bn2bin(bn_invq, tmp_buf);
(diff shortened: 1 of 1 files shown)


jschmidb (approved):
Looks good!

p-steuer:
> As you mention we might want to change this to BN_bn2binpad() once we eliminate support for OpenSSL versions older than 1.1.1. BN_bn2binpad() just looks way more readable.....

Then I suggest to put it in "if opensslversion < xxx ... else ..." already. If you dont do it now, imost likely it wont be remembered at the time support for < 1.1.1 is eleminated.

Approved from a functional perspective.

p-steuer (approved):

hfreude:
Reworked as suggested by Patrick.

ifranzki:
@jschmidb don't you want to include this PR into libica 4.0 ?

PR #73, opened 2021-11-19
Add per key size statistics for AES, RSA and ECC

ifranzki:


src/include/icastats.h
```diff
@@ -167,5 +212,26 @@ char *get_next_usr();
 void stats_reset();
 int delete_all();
 
+static inline int aes_directed_fc_stats_ofs(unsigned int fc)
```
juergenchrist, line 297: Did you check that this function actually gets inlined?

ifranzki: No, but even if not, does it really matter? Its still quite quick....

src/icastats_shared.c
```diff
@@ -53,7 +53,9 @@ int stats_mmap(int user)
 		sprintf(shm_id, "icastats_%d",
 			user == -1 ? geteuid() : (uid_t)user);
 
-		stats_shm_handle = shm_open(shm_id, O_CREAT | O_RDWR,
+		stats_shm_handle = shm_open(shm_id, O_RDWR,  S_IRUSR | S_IWUSR);
+		if (stats_shm_handle == NOT_INITIALIZED)
+			stats_shm_handle = shm_open(shm_id, O_CREAT | O_RDWR,
```
juergenchrist, line 58: If I execute this as root for another user, that user will later not be able to open its shared memory segment since it is created by root with only read and write permissions for user root...

src/include/s390_aes.h
```diff
@@ -140,7 +140,7 @@ static inline int s390_aes_gcm(unsigned int fc, const unsigned char *in_data,
 	if (rc)
 		return rc;
 
-	stats_increment(ICA_STATS_AES_GCM,
+	stats_increment(ICA_STATS_AES_GCM_128 + aes_directed_fc_stats_ofs(fc),
```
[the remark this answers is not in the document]
ifranzki: Well no (although I also first thought that this should/would be the case).
If you follow the caller, the function code is built via `aes_directed_fc()`, just like with any other AES algos.

The code above translates that fc to the KMA-FC:
`rc = s390_aes_gcm_hw(s390_kma_functions[fc].hw_fc, ....`

src/icastats_shared.c
ifranzki: there is this block below, isn't this changing the owner to the user it created it for?

```
		if (user > 0 && geteuid() == 0) {
			if (fchown(stats_shm_handle, user, user) == -1) {
				close(stats_shm_handle);
				return -1;
			}
		}
```

juergenchrist: Yes, missed that in the diff...

ifranzki: BTW: We theoretically have the same problem with OCK statistics. There I did have a fchown as well, but you commented that it would it be required :-) But actually, the code in OCK will never open it for another user, so this is only a theoretical problem there.

src/ica_api.c
```diff
@@ -3639,7 +3639,9 @@ unsigned int ica_aes_cmac_intermediate(const unsigned char *message,
 		       iv);
 
 	if (!rc)
-		stats_increment(ICA_STATS_AES_CMAC, ALGO_HW, ICA_DECRYPT);
+		stats_increment(ICA_STATS_AES_CMAC_128 +
+				aes_directed_fc_stats_ofs(function_code),
```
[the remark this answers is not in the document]
ifranzki: Yes, I thought about that, but didn't want to use it because its not so readable/understandable ....
However, I am open to change it if thats what we want.

jschmidb:
Looks good to me after a first browse through. A minor comment on spelling here:

>  stat: Fix accessing other users statistics shared memory segments
> 
> Trying to access an other users shared memory segment as root with
> shm_open(...., O_CREAT | O_RDWR, S_IRUSR | S_IWUSR) fails if the
> shared memory segment already exists.

Maybe better: Trying to access another user's shared memory segment ...  
I'm going to install and run the code now ...

jschmidb:
ok, after running the code I have two questions:
1. Is there a reason for just printing the json format on the console and not creating an output file? Yes, in scripts this output can be piped ... 
2. I see the comment in the manpage, saying: 

> For the JSON output format, only  the per key size counters are reported.

Is there a reason for this?

ifranzki:
> Is there a reason for just printing the json format on the console and not creating an output file?

Yes, you can pipe it to wherever you want, or redirect in into a file. JSON is mainly meant for further processing by other programs (not by humans), so the output will most likely piped to other programs. And furthermore thats simply the way things are usually done in Unix/Linux. 

>For the JSON output format, only the per key size counters are reported.
>Is there a reason for this?

Yes, again, its meant for further processing by other programs. If they need a sum of all key sizes, or all AES algos, or whatever, they can calculate that easily with the data provided. Providing the sum as part of the JSON output would just increate the amount of data, but not add any additional information.

ifranzki:
> Maybe better: Trying to access another user's shared memory segment ...

Fixed.

jschmidb:
ok, got it, thanks!

jschmidb (approved):
lgtm.

PR #74, opened 2021-12-01
Libica 4.0: Remove sw fallbacks, stage 2

jschmidb:
After introducing libica-cex.so (which has no CPACF support and no software fallbacks), this pr now completes removing sw fallbacks from libica by making -DNO_SW_FALLBACKS the new default for libica.so. RSA key generation is (and ever was) only available via sw (openssl) and is considered to be a "sw feature" and no "sw fallback". Therefore RSA key generation is still available via openssl. Here are the main changes from an external point of view:

- Removing sw fallbacks from libica.so changes the API behavior: applications are now themselves responsible for implementing fallbacks if desired. This API change makes a new major libica version 4.0.0 necessary. 
- Deprecated API functions are now removed, including their tests
- I also did some rework on existing API functions: introduced 'const' to const parms whenever applicable.
- The API function ica_fips_status() was only available when libica was configured and built with ICA_FIPS. This made it impossible for applications to query the fips status from a libica built without ICA_FIPS. The function is now always available.
- icainfo now shows two additional output lines for RSA_KEY_GEN_ME and  RSA_KEY_GEN_CRT.


juergenchrist (approved):

src/fips.c
```diff
@@ -256,6 +256,17 @@ fips_init(void)
...
+	} else {
+		/* kernel fips flag == 0, load default provider in case we are
+		 * running with openssl 3.0 */
+#if OPENSSL_VERSION_PREREQ(3, 0)
+		openssl_provider = OSSL_PROVIDER_load(openssl_libctx, "default");
+		if (openssl_provider == NULL) {
+			syslog(LOG_ERR, "Libica: failed to load default provider\n");
+			fips |= ICA_FIPS_INTEGRITY;
+			return;
+		}
```
ifranzki, line 269: If not compiled against OpenSSL 3.0 this is an empty block... does this produce any compile warnings/errors? 

If so, you could move the `#if OPENSSL_VERSION_PREREQ(3, 0)` before the `} else {` and the `#endif` after the else block.

ifranzki (approved):

ifranzki:
Don't you also want to include [PR 71](https://github.com/opencryptoki/libica/pull/71) into libica 4.0 ?

jschmidb:
> Don't you also want to include PR 71 into libica 4.0 ?

Yes, thanks for the reminder.

PR #75, opened 2021-12-15
RSA: limit RSA key length to 4096

jschmidb:
CEX adapters support RSA up to 4096 bits. Although RSA key generation in libica is done via openssl, and therefore even greater key lengths would be supported, such keys could not be processed on CEX adapters afterwards. With the removal of sw fallbacks this is now a hard restriction.
The check is not so obvious: while the modulus_bit_length parm specifies the desired key length in bits, the public_key.key_length and private_key.key_length specify bytes.

### DIFF
--- a/src/ica_api.c
+++ b/src/ica_api.c
@@ -52,6 +52,8 @@
 
 #define MAX_VERSION_LENGTH 16
 
+#define MAX_RSA_KEY_BITS		4096
+
 #ifndef NO_SW_FALLBACKS
 int ica_fallbacks_enabled = 1;
 #else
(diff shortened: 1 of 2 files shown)


src/ica_api.c
```diff
@@ -1071,8 +1073,10 @@ unsigned int ica_rsa_key_generate_mod_expo(ica_adapter_handle_t adapter_handle,
 	/* Keys should comply with modulus_bit_length */
 	if ((modulus_bit_length + 7) / 8 != public_key->key_length)
 		return EINVAL;
-	/* Minimum length for public exponent is sizeof(unsigned long) */
-	if (public_key->key_length < sizeof(unsigned long))
+	/* Minimum key length is sizeof(unsigned long), max key length is 4096
+	 * because of CEX adapter restriction */
+	if (public_key->key_length < sizeof(unsigned long) ||
+		public_key->key_length > MAX_RSA_KEY_BYTES)
 		return EINVAL;
```
ifranzki, line 1080: Would it make sense to return a specific error code for the case that the RSA key length is not supported by the HW? Similar to the specific return code when an EC curve is not supported. That way the application has a chance to detect the reason for the error.

jschmidb: Good idea ... I'm now returning EPERM in case the modulus bit length is greater than 4096

ifranzki (approved):
lgtm

PR #76, opened 2022-01-05
Fix memory leaks and a recursive loop with OpenSSL 3.0

ifranzki:


jschmidb:
Good findings! RSA is even very old code.
Just one typo in commit  ECC: Avoid recursive loop in ec_key_check() with OpenSSL 3.0:
Use libia's  ....   -> libica's

jschmidb (approved):
lgtm

ifranzki:
Fixed the typo in the commit message. No code changes.

PR #77, opened 2022-01-13
EC: Handle unsupported EC curve in ica_ec_key_new()

ifranzki:
In case of an unsupported curve nid, `privlen_from_nid()` returns -1. The subsequent `calloc()` will fail with a size of -3 (0xfffffffffffffffd).

Also free the already allocated key in case of an error to avoid a memory leak.


jschmidb:
Good catch, lgtm.

jschmidb (approved):

PR #78, opened 2022-01-13
Compute HMAC from installed library

jschmidb:
The HMAC hash was computed from the libica in the build tree, but
the runtime check is run against the installed libica and those 2
files may be different. E.g. if the runtime one has debuginfo stripped
(and placed into a separate file), the hashes are different.
This commit introduces a new make target: fipsinstall, which creates
the HMAC files in the install directory.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -17,11 +17,8 @@ coverage: check
 	cd ${top_builddir}/src && gcov .libs/*.gcda
 
 if ICA_FIPS
-install-data-hook:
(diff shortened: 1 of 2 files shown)


sharkcz (approved):
LGTM

sharkcz:
And if I see right in the build log, then the HMAC files are created at the start of `make check` ...

ifranzki:
The tests do a `export LD_LIBRARY_PATH=${builddir}/../src/.libs/` to use the library from the build tree.

And yes, I see that src/Makefile.am also generates the HMAC files for the library in the build tree, so then all is good.

jschmidb:
Ingo, thanks for commenting!

PR #79, opened 2022-01-24
Fix compile warnings and memory leaks

ifranzki:


jschmidb:
lgtm

jschmidb (approved):

ifranzki:
Added one more commit, please re-review.

src/Makefile.am
```diff
@@ -90,8 +90,6 @@ icastats_SOURCES = icastats.c icastats_shared.c include/icastats.h
 # internal tests
 
 if ICA_INTERNAL_TESTS
-bin_PROGRAMS += internal_tests/ec_internal_test
-
```
jschmidb, line 94: Good catch! Yes, we don't want the internal tests to be installed. 
However, when I test this change, I'm getting:
```
# ./bootstrap.sh
...
src/Makefile.am:103: warning: variable 'internal_tests_ec_internal_test_SOURCES' is defined but no program or
src/Makefile.am:103: library has 'internal_tests_ec_internal_test' as canonical name (possible typo)
src/Makefile.am:102: warning: variable 'internal_tests_ec_internal_test_LDADD' is defined but no program or
src/Makefile.am:102: library has 'internal_tests_ec_internal_test' as canonical name (possible typo)
```
Besides of this warning, everything seems to be fine and make install does no longer install the internal tests. I googled for a solution of this warning and indeed found some hints, but nothing really worked here.

ifranzki: Arg, didn't see that. And Travis is also failing. 
Let me look, this needs to be fixed.....

ifranzki: noinst_PROGRAMS is what we need. Will push shortly....

ifranzki:
Should be OK now.

jschmidb:
yep, runs perfect now.

PR #83, opened 2022-03-31
OpenSSL 3.0: Cleanup OpenSSL library context during OpenSSL cleanup

ifranzki:
Usually libica's own library context is freed in the library destructor when the library is unloaded (i.e. during exit handlers).

OpenSSL also performs its own cleanup in exit handlers, and it may happen that OpenSSL cleanup is performed before the library destructors are called. This may cause crashes when libica's library context has already been freed by OpenSSL cleanup, but the library destructor tries to free it a second time. This causes a double free, and very likely a crash.

Register an OpenSSL cleanup handler to clean up the library context before OpenSSL performs its own cleanup.

Similar fix as for Opencryptoki in https://github.com/opencryptoki/opencryptoki/pull/529

### DIFF
--- a/src/init.c
+++ b/src/init.c
@@ -65,6 +65,18 @@ void end_sigill_section(struct sigaction *oldact, sigset_t *oldset)
 	sigprocmask(SIG_SETMASK, oldset, NULL);
 }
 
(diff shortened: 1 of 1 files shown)


jschmidb (approved):
lgtm

PR #84, opened 2022-04-05
Openssl cleanup - next try

ifranzki:


src/init.c
```diff
@@ -65,18 +65,6 @@ void end_sigill_section(struct sigaction *oldact, sigset_t *oldset)
 	sigprocmask(SIG_SETMASK, oldset, NULL);
 }
 
-#if OPENSSL_VERSION_PREREQ(3, 0)
-static void openssl_cleanup()
```
juergenchrist, line 69: How about making this a public API function and document that the user should call this before dlclose()?  That seems to be the cleanest solution.  Would also be an API addition (thus a minor release).

ifranzki: Would be OK for me. We should probably use a different name then, like ica_openssl_cleanup().....
And we will need to have this function independent on the OpenSSL version. For < OpenSS 3.0 this would just be a no-op.

Not sure if users will really call that function then, but at least there is a possibility to do it right then.

@jschmidb What do you think?

jschmidb: That would certainly be ok with me. Would opencryptoki then use this new API function?

ifranzki: Well, this would then mean a version dependency... not sure how to deal with tis best?
So I guess sooner or later it would (as well as the IBMCA engine I guess)

jschmidb (approved):

ifranzki:
Added one commit to introduce an external ica_cleanup function.

To be 100% correct, we would need to call this function also in all the testcases.....

libica.map
```diff
@@ -166,3 +166,9 @@ LIBICA_3.6.0 {
 	ica_ed448_ctx_del;
     local: *;
 } LIBICA_3.5.0;
+
+LIBICA_4.0.2 {
```
ifranzki, line 170: This might need to be updated whatever the next release will be....

src/init.c
juergenchrist: For the IBMCA engine, that is easy.  Lookup the symbol via dlsym.  If it is there, call it before dlclose.  If not, do not call it.  For the provider, this might be trickier.  But that is what autoconf is for...

juergenchrist:
> To be 100% correct, we would need to call this function also in all the testcases.....

Well, if they directly exit and we do not perform leak checking, we can still get away with it.  Only tests that actually continue after a dlclose or do leak checking might be a problem.

juergenchrist (approved):

jschmidb:
Just looking at icastats ... no dependency to openssl there, so no hit. Regarding icainfo: when exiting there, all loaded libs are automatically unloaded/freed. Why do you call ica_cleanup() there?

ifranzki:
> Regarding icainfo: when exiting there, all loaded libs are automatically unloaded/freed. Why do you call ica_cleanup() there?

Because in the library constructor the library context has been allocated and provider has been loaded. So we need to unload the provider and free the context before exiting in any case. No matter how and who loads the library.

juergenchrist:
Okay, that discussion going in the wrong direction seems to be my fault.  All I wanted to say is that I do not see the missing call to ica_cleanup in the test cases as a blocker for this PR.  Just wanted to mention that *if* we ever want to use leak checking (either via msan/lsan or - if ever supported - via valgrind), then this becomes a problem.  But in general, everyone who dlopen()s libica should call this function to prevent leaks.

ifranzki:
> everyone who dlopen()s libica should call this function to prevent leaks.

I would even say everyone who USES libica (no matter how) should call ica_cleanup(). 

But I agree, its not critical if we leak the context and provider, as  longs as we don't use leak checking.

jschmidb:
Good .... so the only point left is using the correct version/release number in all affected files. Ingo changed it in libica.map to 4.0.2 and I would leave this for now. With the next release upgrade that will be necessary because of the API change, I will use the correct release number in all affected files.

jschmidb (approved):

juergenchrist:
The release should be 4.1.0 since this is not a bugfix on an existing API function, but an addition to the API.

PR #85, opened 2022-04-19
Fix memory leak at library unload by uninstantiating global prng instance

ifranzki:
When built in non-FIPS mode, s390_prng_init() initializes a global PRNG instance in the library constructor, which must also be freed in the library destructor. Otherwise it leaks 64 bytes (direct leak) plus 240 bytes (indirect leak) when unloading the library.

### DIFF
--- a/src/include/s390_prng.h
+++ b/src/include/s390_prng.h
@@ -16,5 +16,6 @@
 
 int s390_prng_init(void);
 int s390_prng(unsigned char *output_data, unsigned int output_length);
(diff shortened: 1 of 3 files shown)


jschmidb:
Yep, that's a leak. How did you find it? Did you use a tool?

ifranzki:
> How did you find it?

The address sanitizer shows memory leaks.
But you see this one only when you explicitly unload the library.

jschmidb (approved):

PR #86, opened 2022-05-16
ECC: Signature verify may also return 12 / 769 for bad signatures

ifranzki:
According to the description of CCA verb "Digital Signature Verify" (CSNDDSV) in "Secure Key Solution with the Common Cryptographic Architecture Application Programmer's Guide Version 7.1" the signature verify request may also return 12 / 769 for bad signatures, in addition to 4 / 429. Treat both return codes as bad signature indication.

### DIFF
--- a/src/s390_ecc.c
+++ b/src/s390_ecc.c
@@ -1891,6 +1891,13 @@ unsigned int ecdsa_verify_hw(ica_adapter_handle_t adapter_handle,
 		goto ret;
 	}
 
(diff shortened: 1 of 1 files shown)


jschmidb:
Hmm ... I don't have the CCA 7.1 book here, but the 7.3 book says: 

> For appropriate failure cases, if the value of CSU_EC_CHECKCURVE is 1, then the new return/reason code
> 12/769 is returned. Otherwise, 4/429 is returned.

Can you check again if it's 768 or 769?

jschmidb (approved):

PR #87, opened 2022-05-18
RSA: Retry RSA key gen if it fails with OpenSSL 3.0

ifranzki:
In OpenSSL 3.0 the RSA key gen algorithm has been changed and can now fail to generate a key. Retry up to 10 times in such a case.


jschmidb (approved):
lgtm

PR #88, opened 2022-07-20
v5: reduce the number of open file descriptors

holger-dengler:
This PR contains a set of commits to reduce the number of open file descriptors and a few other cleanups.

The cleanup covers the following:
- Patch 1/7: fix the autotools clean targets. Now all generated files will be removed by these targets
- Patch 2/7: add user support in stats_munmap()
- Patch 3-4/7: increase readability of stats_mmap() and remove redundant code

Openssh has a hard limit of open file descriptors, while initiating/handling a connection. If openssh is using openssl with a configured libica engine, this hard limit to only 6 open file descriptors may hurt. 
Patch 5/7 closes the shm fd right after the mmap(), which is common practice since a very long time. Just for the record: the mapped shm segment can be used, even if the file descriptor has been closed. This change also removes the requirements of a global file descriptor.  
Patch 6/7 closes the syslog related file descriptors. This is a brute-force solution, because it closes the descriptors each time a syslog has been called. But it works and while syslog is only used during initialization, it should not hurt. 
Patch 7/7 move the known HMAC load to a separate function. The added comparison also reduces the attack surface.

### DIFF
--- a/src/fips.c
+++ b/src/fips.c
@@ -1085,7 +1085,7 @@ _err_:
 }
 
 static int
-des3_cbc_cs_kat(void){
+des3_cbc_cs_kat(void) {
 	const struct des3_cbc_cs_tv *tv;
 	size_t i;
 	unsigned char iv[DES3_BLKSIZE], *out;
--- a/src/fips.c
+++ b/src/fips.c
@@ -360,6 +360,37 @@ end:
 	return rc;
 }
 
+static int load_known_hmac(const char *path, unsigned char **hmac, long *hmaclen)
+{
+	int rc = -1;
+	FILE *fp;
+	char *known_hmac_str = NULL;
+	char *hmacpath, *p;
+	size_t n;
+
+	hmacpath = make_hmac_path(path);
+	if (hmacpath == NULL)
+		return rc;
+
+	if ((fp = fopen(hmacpath, "r")) == NULL)
+		return rc;
+
+	if (getline(&known_hmac_str, &n, fp) <= 0)
+		goto end;
+
+	if ((p = strchr(known_hmac_str, '\n')) != NULL)
+		*p = '\0';
+
+	*hmac = OPENSSL_hexstr2buf(known_hmac_str, hmaclen);
+	rc = 0;
+end:
+	fclose(fp);
+	free(known_hmac_str);
+	free(hmacpath);
+
+	return rc;
+}
+
 /**
  * Performs the FIPS check.
  *
@@ -369,49 +400,27 @@ end:
 static int FIPSCHECK_verify(const char *path)
 {
 	int rc = 0;
-	FILE *fp;
 	unsigned char *known_hmac = NULL;
-	long hmaclen;
-	char *hmacpath, *p;
-	char *known_hmac_str = NULL;
-	size_t n, computed_hmac_len;
+	long known_hmac_len;
 	void *computed_hmac = NULL;
+	size_t computed_hmac_len;
 
-	hmacpath = make_hmac_path(path);
-	if (hmacpath == NULL)
-		return 0;
-
-	fp = fopen(hmacpath, "r");
-	if (fp == NULL) {
+	if (load_known_hmac(path, &known_hmac,  &known_hmac_len) != 0)
 		goto end;
-	}
-
-	if (getline(&known_hmac_str, &n, fp) <= 0)
-		goto end;
-
-	if ((p = strchr(known_hmac_str, '\n')) != NULL)
-		*p = '\0';
-
-	known_hmac = OPENSSL_hexstr2buf(known_hmac_str, &hmaclen);
 
 	if (compute_file_hmac(path, &computed_hmac, &computed_hmac_len) != 0)
+		goto end;
+
+	if ((size_t)known_hmac_len != computed_hmac_len)
 		goto end;
 
 	if (memcmp(computed_hmac, known_hmac, computed_hmac_len) != 0)
 		goto end;
 
 	rc = 1;
-
 end:
-
 	free(computed_hmac);
-	free(known_hmac_str);
-	free(hmacpath);
-
 	OPENSSL_free(known_hmac);
-
-	if (fp)
-		fclose(fp);
 
 	return rc;
 }
--- a/src/icastats.c
+++ b/src/icastats.c
@@ -291,7 +291,7 @@ int main(int argc, char *argv[])
 		return EXIT_SUCCESS;
 	} else if(delete){
 		stats_mmap(user);
-		stats_munmap(SHM_DESTROY);
+		stats_munmap(user, SHM_DESTROY);
 		return EXIT_SUCCESS;
 	}
 	if(all){
--- a/src/icastats.c
+++ b/src/icastats.c
@@ -88,7 +88,7 @@ void print_stats(stats_entry_t *stats, int key_sizes)
 		if (!key_sizes && strncmp(STATS_DESC[i], "- ", 2) == 0)
 			continue;
 
-		if(i <= ICA_STATS_RSA_CRT_4096) {
+		if (i <= ICA_STATS_RSA_CRT_4096) {
 			printf(" %14s |        %*lu          |         %*lu\n",
 			       STATS_DESC[i],
 			       CELL_SIZE,
@@ -106,8 +106,7 @@ void print_stats(stats_entry_t *stats, int key_sizes)
 			       stats[i].enc.sw,
 			       CELL_SIZE,
 			       stats[i].dec.sw);
-
-	       }
+		}
 	}
 }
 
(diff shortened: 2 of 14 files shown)


src/icastats_shared.c
```diff
@@ -46,74 +44,77 @@ volatile int stats_shm_handle = NOT_INITIALIZED;
...
 	munmap(stats, STATS_SHM_SIZE);
-	close(stats_shm_handle);
-	stats_shm_handle = NOT_INITIALIZED;
+	stats = NULL;
+
+	if(unlink == SHM_DESTROY) {
+		char shm_id[NAME_LENGHT];
+
+		sprintf(shm_id, "icastats_%d",
+			user == -1 ? geteuid() : (uid_t)user);
```
juergenchrist, line 114: Hopefully we are not setuid when calling this...  But in that case, simply do not call this function with -1 for `user` and `SHM_DESTROY` for unlink.

juergenchrist (approved):
The syslog stuff is still a little messed up.  But that needs to be cleared somewhere else.  I see us logging at runtime only in the RNG stuff.  So I am wondering if we should go the 
```
openlog()
fips_tests_in_init
closelog()
ica_open_adapter
openlog
```
way.  Would save some `closelog` calls.  But I am fine with the current version, too.

src/icastats_shared.c
holger-dengler: The more I think about the syslog stuff, the more I come to the conclusion that we should remove all syslog() calls outside of the initialization. It is the responsibility of the caller of a library, to print/log/syslog errors. The library should only provide a possibility to detect a problem (e.g. by providing a error code). After removing all syslog() in rng_init() and drbg/set_error_state(), we can do a closelog() at the end of the initialization.

juergenchrist: I think that comment went to the wrong place.  Here we do nothing with syslog/closelog but re-create the name of the segment.

holger-dengler: uups

holger-dengler: Anyhow: I'll rework the syslog stuff and provide a v2.

holger-dengler:
The update contains some reworks.

- syslog: the closelog() is now only done at the end of the library initialization. 
- tests: a test has been added for checking the value of the adapter handle. It should not be greater than 3.
- rng_gen(): the exit(1) has been replaced by a proper error handling. The syslog() has been removed.

src/rng.c
```diff
@@ -56,13 +56,13 @@ void rng_gen(unsigned char *buf, size_t buflen)
...
-				return;
+				return 0;
 		}
 	}
 
-	syslog(LOG_ERR, "Libica internal RNG error..");
-	fprintf(stderr, "Libica internal RNG error.");
-	exit(1);
+	/* unable to get any random data. cleanup and return with error */
+	memset(buf, 0, buflen);
```
holger-dengler, line 64: I found a missing #include <string.h>, so the memset() usage causes a warning. I'll fix that issue.

src/fips.c
```diff
@@ -360,59 +360,68 @@ static int compute_file_hmac(const char *path, void **buf, size_t *hmaclen)
...
+	void *computed_hmac = NULL;
+	size_t computed_hmac_len;
+
+	if (load_known_hmac(path, &known_hmac,  &known_hmac_len) != 0)
+		goto end;
 
 	if (compute_file_hmac(path, &computed_hmac, &computed_hmac_len) != 0)
 		goto end;
 
+	if (known_hmac_len != computed_hmac_len)
```
holger-dengler, line 414: cast to size_t is missing. I'll fix this for the next push.

holger-dengler:
I pushed the next version to the tip branch.

src/icastats.c
```diff
@@ -291,7 +291,7 @@ int main(int argc, char *argv[])
 		return EXIT_SUCCESS;
 	} else if(delete){
 		stats_mmap(user);
-		stats_munmap(SHM_DESTROY);
+		stats_munmap(user, SHM_DESTROY);
 		return EXIT_SUCCESS;
 	}
 	if(all){
```
holger-dengler, line 297: I think, that this also fixes a bug. 

`stats_mmap(user)` will create a shm-segment with the filename `icastats_<user>` Without the fix, the following `stats_munmap(SHM_DESTROY)` will unlink the shm-segment with the filename `icastats_<euid>`. This is (IMHO) an error in cases, where `user != euid`. Or did I missed something?

test/adapter_handle_test.c
```diff
@@ -0,0 +1,41 @@
+/* This program is released under the Common Public License V1.0
+ *
+ * You should have received a copy of Common Public License V1.0 along with
+ * with this program.
+ */
+
+/* Copyright IBM Corp. 2022 */
+
+#include <stdio.h>
+#include <sys/errno.h>
```
juergenchrist, line 10: Why not simply `#include <errno.h>`?

test/adapter_handle_test.c
```diff
@@ -0,0 +1,41 @@
...
+
+#include "ica_api.h"
+#include "testcase.h"
+
+extern int errno;
```
juergenchrist, line 15: This should not be needed at all.

src/rng.c
```diff
@@ -38,7 +39,7 @@ void rng_init(void)
 			     sizeof("INTERNAL INSTANCE"));
 }
 
-void rng_gen(unsigned char *buf, size_t buflen)
+int rng_gen(unsigned char *buf, size_t buflen)
```
juergenchrist, line 42: The EC and ED key gens use this function, but do not check the return value.  But maybe first discuss with Joerg whether we want to fix this or remove the key gens and simply use OpenSSL.  This discussion is part of an ongoing effort to make libica FIPS 140-3 compliant.

test/adapter_handle_test.c
holger-dengler: hm, good question. Seems like a copy&paste leftover. I'll fix it.

test/adapter_handle_test.c
holger-dengler: ok, I'll fix it.

src/rng.c
holger-dengler: This commit also changes the implementation of `ica_[x|ed][25519|448]_key_gen()`, where `rng_gen()` is called. All these functions already return `-1` is `fips_check() != 0`. I just use not the same return value, if `rng_gen()` fails. This in in my opinion the correct solution and much better than doing a exit(1)!

holger-dengler:
This version contains:
- fixes for travis
- test cleanup

src/icastats_shared.c
```diff
@@ -46,74 +44,77 @@ volatile int stats_shm_handle = NOT_INITIALIZED;
...
 
-		stats = (stats_entry_t *) mmap(NULL, STATS_SHM_SIZE, PROT_READ |
-						 PROT_WRITE, MAP_SHARED,
-						 stats_shm_handle, 0);
-		if (stats == MAP_FAILED){
-			close(stats_shm_handle);
-			stats = NULL;
-			return -1;
-		}
+	if (ftruncate(stats_shm_handle, STATS_SHM_SIZE) == -1)
```
juergenchrist, line 71: We should not do that if we did not generate the segment.  What if - for whatever reason - two different versions of libica are running in the system with different `STATS_SHM_SIZE` values?

holger-dengler: This might be a valid point, but please keep in mind, that my change just removed one level of indention and the curly brackets. This is the only reason, why the line appeared in this commit.

holger-dengler: A fix for this potential issue should anyhow go into a separate commit.

test/adapter_handle_test.c
```diff
@@ -0,0 +1,37 @@
...
+	set_verbosity(argc, argv);
+
+	rc = ica_open_adapter(&adapter_handle);
+	if (rc != 0) {
+		V_(perror("ica_open_adapter failed"));
+		return TEST_FAIL;
+	}
+
+	if (adapter_handle > 3) {
+		V_(printf("ica_open_adapter: file descriptor value should "
```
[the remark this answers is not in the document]
holger-dengler: If >80 character lines are acceted, fine with me.

src/icastats_shared.c
```diff
@@ -59,41 +58,40 @@ int stats_mmap(int user)
...
 
 	if (stat_buf.st_size != STATS_SHM_SIZE)
 		memset(stats, 0, STATS_SHM_SIZE);
 
+	close(stats_shm_handle);
 	return 0;
+
+err:
+	close(stats_shm_handle);
+	return -1;
```
holger-dengler, line 91: Use rc as return value and merge err label with good path. I will fix this in the next version of the series.

holger-dengler:
Wording in commit message of patch 6/8 will be fixed.

test/adapter_handle_test.c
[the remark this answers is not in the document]
holger-dengler: Will be fixed in next version.

holger-dengler:
A new version has been pushed, which contains the following changes:
- new commit with space/blankline fixes
- various review comments
- fixed wording in commit message

jschmidb (approved):
As discussed offline, I'm now going to merge the pr. 
Thanks for fixing and reviewing!

PR #89, opened 2022-08-23
Fips140-3 compliance

jschmidb:
This pull request provides patches to make libica FIPS 140-3 compliant. It upgrades libica to new minor level 4.1.0.


src/s390_crypto.c
```diff
@@ -695,6 +724,7 @@ int s390_get_functionlist(libica_func_list_element *pmech_list,
 		pmech_list[x].mech_mode_id != RSA_KEY_GEN_CRT)
 		pmech_list[x].flags &= ~ICA_FLAG_SW;
 #endif
+
```
ifranzki, line 727: Empty line added. Not sure if this is intentional.....

src/s390_crypto.c
```diff
@@ -675,6 +666,44 @@ int s390_get_functionlist(libica_func_list_element *pmech_list,
...
+		}
+
+		/* ECDSA/ECDH only via CPACF in FIPS 140-3 mode */
+		switch (pmech_list[x].mech_mode_id) {
+		case EC_DH:
+		case EC_DSA_SIGN:
+		case EC_DSA_VERIFY:
+		case EC_KGEN:
+			pmech_list[x].flags &= ~ICA_FLAG_DHW;
+			pmech_list[x].flags &= ~ICA_FLAG_SW;
```
ifranzki, line 695: So here you essentially state that in Fips mode EC operations as well as EC key gen only operate on CPACF (thus no DHW and no SW). 

1. In the next commits you force the EC key gen to happen by OpenSSL, this is SW. So at least for EC Keygen, it thus should be only the SW flag, but no others.
2. Are you actually restricting the EC operations to happen only on CPACF? I don't seem to find that code in any upcoming commit. I only see that you perform selftests only on CPACF (because you need a deterministic K value). But for regular, API driven EC requests it goes the regular path, and thus may choose CPACF, card or SW as available.

So please double check if the flag setting is correct here.

src/s390_crypto.c
```diff
@@ -675,6 +666,44 @@ int s390_get_functionlist(libica_func_list_element *pmech_list,
...
+			if (pmech_list[x].flags) {
+				pmech_list[x].property &= ~ICA_PROPERTY_EC_BP;
+				pmech_list[x].property &= ~ICA_PROPERTY_EC_ED;
```
ifranzki, line 698: Similar as above, are you actually restricting the API driven EC requests to not use BP and ED curves?

src/ica_api.c
```diff
@@ -1145,6 +1145,11 @@ unsigned int ica_rsa_mod_expo(ica_adapter_handle_t adapter_handle,
 	if (rsa_key->key_length * 8 > MAX_RSA_KEY_BITS)
 		return EPERM;
 
+#ifdef ICA_FIPS
+	if ((fips & ICA_FIPS_MODE) && (rsa_key->key_length * 8 < 2048))
```
ifranzki, line 1149: Brackets around `rsa_key->key_length * 8 < 2048` are not needed.
Same at the other 3 similar places further down.

src/icainfo.c
```diff
@@ -127,6 +127,74 @@ static int online_cca_card(void)
...
+
+/**
+ * Print out the minimum and maximum RSA key length. The maximum length is
+ * restricted to 4096 bits by crypto cards. The minimum accepted length in
+ * libica is 57 bits, but the available min length depends on the openssl
+ * version and fips mode.
+ */
+void print_rsa(void)
+{
+	int keylen_array[] = { 57, 512, 1024, 2048, 4096 };
```
ifranzki, line 175: does anyone really care about an RSA key size of 57 bits ????
Even 512 bits is questionable.

src/ica_api.c
jschmidb: Done.

src/icainfo.c
jschmidb: Yes indeed. I guess this is very old code in libica, but libica (and in fact openssl!) still support RSA with 57 bits. I would leave it up to applications whether to use such small key sizes, but found it worth to display the lower and upper boundaries. Note that the icainfo -r output just displays what openssl really allows. If openssl is configured to restrict small key sizes (e.g. via the fips provider/mode) then this output reflects this.

src/s390_crypto.c
jschmidb: Empty line is removed with the fix for EC keygen sw flag, see comment below.

src/s390_crypto.c
jschmidb: See comment above, yes I think so, but will double-check.

src/fips.c
```diff
@@ -1003,6 +1005,147 @@ aes_cmac_kat(void)
...
+		}
+		ica_ec_key_free(eckey);
+	}
+
+	return 0;
+
+_err_:
+	syslog(LOG_ERR, "Libica ECDSA test failed.");
+	return 1;
+}
```
holger-dengler, line 1147: Same here for `eckey`.

src/s390_crypto.c
ifranzki: I don't see any fips checks (other than the general fips state is bad checks) that would prevent anyone from using certain curves when running in fips mode.... but maybe I am just not seeing it....
I also don't see any restriction on HW or SW usage when fips in on...... So even if OpenSSL would restrict certain curves when in fips mode, once it goes to the card or CPACF, then there is no restriction it seems.

src/fips.c
```diff
@@ -1003,6 +1005,147 @@ aes_cmac_kat(void)
...
+		ica_ec_key_free(eckey_B);
+	}
+
+	ica_close_adapter(ah);
+	return 0;
+
+_err_:
+	ica_close_adapter(ah);
+	syslog(LOG_ERR, "Libica ECDH test failed.");
+	return 1;
```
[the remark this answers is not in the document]
jschmidb: Thanks! Will fix.

src/fips.c
jschmidb: will fix.

jschmidb:
Just pushed the fixes related to your comments so far:

- @ifranzki Refer to new commit "Restrict EC/ED/X curves in fips mode" for API restrictions. I also updated commit "handle fips mode in function list" to set the flags correctly for EC keygen. icainfo now correctly displays only a 'yes' in the sw column. 
`    EC Keygen |     no     |     no     |    yes`
For non-fips the output still shows:
`     EC Keygen |    yes     |    yes     |      -`
which is what we want.
- There was a comment via email from ATSEC, saying "all temporary buffers used to calculate the integrity test must be zeroized". Refer to new commit "clear temp buffers used for integrity test" for this.

jschmidb:
Just saw that the Travis build fails with the new updates, because the eddsa_test and x_test fail. This is not the case on my test system, will check ...

src/ica_api.c
```diff
@@ -1350,17 +1350,17 @@ int ica_ec_key_init(const unsigned char *X, const unsigned char *Y,
...
 
+	/* check for obvious errors in parms */
+	if (key == NULL)
+		return EINVAL;
+
 #ifdef ICA_FIPS
 	if (fips >> 1)
 		return EACCES;
-	if (!curve_supported_via_openssl(key->nid))
+	if (!curve_supported_via_openssl(key->nid) || !curve_supported_via_cpacf(key->nid))
```
holger-dengler, line 1370: I know, that `fips >> 1` has been checked in the line before, but maybe you can use `check_fips()` here and in all places below.

src/ica_api.c
```diff
@@ -1669,10 +1676,14 @@ void ica_ec_key_free(ICA_EC_KEY *key)
 	free(key);
 }
 
-static inline int check_fips(void)
+static inline int check_fips(unsigned int nid)
```
holger-dengler, line 1705: I would prefer a function name, which better describes, what this function is checking, e.g. `fips_restricted()` or similar. While at it, it might be moved to fips.c?

src/ica_api.c
```diff
@@ -1745,7 +1756,7 @@ int ica_x25519_key_set(ICA_X25519_CTX *ctx,
 	UNUSED(pub);
 	return EPERM;
 #else
-	if (check_fips() || !msa9_switch || ctx == NULL)
+	if (check_fips(NID_X25519) || !msa9_switch || ctx == NULL)
 		return -1;
```
ifranzki, line 1786: I always wondered why the Edwards functions in most cases return -1 instead of a useful errno value, like the other functions do.

ifranzki:
> @ifranzki Refer to new commit "Restrict EC/ED/X curves in fips mode" for API restrictions. 

This looks good.

However, I still wonder if you would not also need some restrictions for EC curves:
First, restrict the curves that are allowed with FIPS (if not already done).
Second: You run the EC selftest using CPACF only. Is it OK for FIPS to perform the API calls for EC operations with another code paths than the one that was self-tested? I.e. wouldn't you also have to use CPACF for all EC operations when in fips mode? If you allow to perform EC operations with the card and/or SW, then it runs code paths that have not been self tested. I don't know the FIPS requirements good enough to answer that.....

src/fips.c
```diff
@@ -379,6 +380,7 @@ static int load_known_hmac(const char *path, unsigned char **hmac, long *hmaclen
 	rc = 0;
 end:
 	fclose(fp);
+	OPENSSL_cleanse(known_hmac_str, strlen(known_hmac_str));
```
holger-dengler, line 383: `known_hmac` and `known_hmac_str` are just read buffers, and are not involved in any hash calculation. Strictly speaking it should not be seccessary to zerorize them as well. But on the other side, it won't hurt to clean them as well. So I would leave it as is.

src/ica_api.c
jschmidb: Yes. I could change these returns to something like EPERM or EIO. But then we would have to distinguish between "no msa9", which would be EPERM or EIO, and ctx == NULL, which would be more like EINVAL. Unfortunately we externally document the rc = -1 in ica_api.h and the libica book. So I would do that later with a cleanup patch. It's noted on my todo list ;-)

jschmidb:
> However, I still wonder if you would not also need some restrictions for EC curves:
> First, restrict the curves that are allowed with FIPS (if not already done).

Yes, the idea here was to check with curve_supported_via_openssl(nid) to filter out curves that are not supported by openssl (in fips mode)

> Second: You run the EC selftest using CPACF only. Is it OK for FIPS to perform the API calls for EC operations with another code paths than the one that was self-tested? I.e. wouldn't you also have to use CPACF for all EC operations when in fips mode? If you allow to perform EC operations with the card and/or SW, then it runs code paths that have not been self tested. I don't know the FIPS requirements good enough to answer that.....

Yes, the idea here was to check with curve_supported_via_cpacf(nid) to only allow the CPACF path, which is self-tested.

The full check for fips mode is:
```
#ifdef ICA_FIPS
	if (fips >> 1)
		return EACCES;
	if (!curve_supported_via_openssl(privkey->nid) ||
		!curve_supported_via_cpacf(privkey->nid))
		return EPERM;
#endif /* ICA_FIPS */
```
But when thinking about this, I'm realizing that for ED/X this check works, but could be simpler: as of now we restrict ED/X in fips mode, simply because they are not approved by NIST, not because openssl does not support them in fips mode. 

So the routine:
```
static inline int check_fips(unsigned int nid)
{
#ifdef ICA_FIPS
	if (fips >> 1 || !curve_supported_via_openssl(nid) ||
		!curve_supported_via_cpacf(nid))
		return 1;
	else
		return 0;
#else
	return 0;
#endif
}
```
could be just like:
```
#ifdef ICA_FIPS
	return 1;    <- even independent of the global 'fips' variable
#else
	return 0;
#endif
}
```

src/ica_api.c
ifranzki: ... and this might be considered as an API breakage ...

ifranzki:
> Yes, the idea here was to check with curve_supported_via_cpacf(nid) to only allow the CPACF path, which is self-tested.

OK, makes some sense, but for example in ica_ecdsa_sign() I see that you added:
```
if (!curve_supported_via_openssl(privkey->nid) ||
		!curve_supported_via_cpacf(privkey->nid))
		return EPERM;
```
So you reject all curves not supported by OpenSSL nor by CPACF. So far so good.

But with that, the operation could still be performed by the card (e.g. if icapath = 1, and ica_offload_enabled = 1, and the curve is also supported by the card), no ?

I guess you somehow need to force it to use CPACF (or maybe OpenSSL if we assume the OpenSSL is FIPS tested as well), but never the card.

jschmidb:
Yes, you are right. I think I see the gap: if we are e.g. running on a pre-z15 machine with a CCA card, the icapath is indeed 1 (because sw callbacks are disabled), but inside ecdsa_sign_hw() the CEX path would be taken, because msa9 is not available. Or, as you say, if ica_offloads are enforced. I'm adding a fix to prevent ecdsa_sign_hw/verify_hw from going through CEX adapters.

holger-dengler (approved):
Looks good to me.

ifranzki (approved):
looks good now.

jschmidb:
I removed the "libica 4.1.0" version patch and will now merge. Next steps are merging Holger's pr and providing a new patch for EC pairwise key / public key checking as required by SP800-56Arev3. The version patch then comes on top again when there are no more remaining questions about FIPS 140-3.

PR #90, opened 2022-08-25
fips140-3: additional changes

holger-dengler:
This series contains 2 cleanup patches, which should be merged into the current fips140-3 branch. The first patch separates the pkey handling, which also makes the memory cleanup a bit more handy.
The second commit is the main purpose of the PR. It replaces the reads with an mmap-handing in the library integrity check. The main reason for this is not only the increased performance, but it allows to get rid of the huge read buffer on the stack.


src/fips.c
```diff
@@ -310,17 +314,36 @@ static EVP_PKEY *get_pkey(void)
...
+	if (fd < 0)
+		goto end;
+
+	if (fstat(fd, statbuf) < 0)
+		goto end_close;
+
+	fp = mmap(0, statbuf->st_size, PROT_READ, MAP_SHARED, fd, 0);
+	if (fp == MAP_FAILED)
+		fp = NULL;
+end_close:
```
ifranzki, line 332: isn't this always closing the file ? I guess a `goto end;` is missing in the good case-

holger-dengler: In general, if a file is accessed via mmap, the fd can be closed as soon as the file is successfully mmap-ed in address space of the process. This is the common file-mmap processing in unix/linux projects.

So: We're always closing the file and do not need a `goto end;` for the good case.

ifranzki: Ah sure, my fault. I thought you are returning the fd, but you are returning the mapped pointer.
All is good. Sorry for the noise.

holger-dengler: np, I also thought so during the review before sending the PR... :) Maybe I should rename fp to fdata in this function as well.

ifranzki: > Maybe I should rename fp to fdata ...

that would help.

holger-dengler: ok, I will change it.

src/fips.c
```diff
@@ -330,41 +353,39 @@ static int compute_file_hmac(const char *path, void **buf, size_t *hmaclen)
...
-		if (EVP_DigestSignUpdate(mdctx, rbuf, len) <= 0) {
-			goto end;
-		}
-	}
+	if (EVP_DigestSignUpdate(mdctx, fdata, fdata_stat.st_size) != 1)
+		goto end;
 
-	hlen = sizeof(rbuf);
-	if (EVP_DigestSignFinal(mdctx, rbuf, &hlen) <= 0)
+	if (EVP_DigestSignFinal(mdctx, tmp, &tmp_len) != 1)
```
ifranzki, line 366: You may actually use EVP_DigestSign() instead of EVP_DigestSignUpdate() followed by EVP_DigestSignFinal().

holger-dengler: Ok, I'll change it.

holger-dengler:
changes since previous version:
- rename fp to ptr to increase the readability of mmap_file()
- use EVP_DigestSign() instead of EVP_DigestSignUpdate()/EVP_DigestSignFinal()

ifranzki (approved):
When rebased to the current master, the remaining 2 commits look good now.

holger-dengler:
The branch has been rebased to the opencryptoki:master.

PR #91, opened 2022-08-31
compile warning fixes

holger-dengler:
This series contains fixes for some compile warnings:
- use of uninitialized variable
- discard of const qualifier

### DIFF
--- a/src/fips.c
+++ b/src/fips.c
@@ -1100,14 +1100,11 @@ ecdh_kat()
 		tv = &ECDH_KAT_TV[i];
 
 		eckey_A = ica_ec_key_new(tv->nid, &privlen);
-		if (!eckey_A)
+		eckey_B = ica_ec_key_new(tv->nid, &privlen);
+		if (!eckey_A || !eckey_B)
 			goto _err_;
 
 		if (ica_ec_key_init(tv->xa, tv->ya, tv->da, eckey_A))
-			goto _err_;
-
-		eckey_B = ica_ec_key_new(tv->nid, &privlen);
-		if (!eckey_B)
 			goto _err_;
 
 		if (ica_ec_key_init(tv->xb, tv->yb, tv->db, eckey_B))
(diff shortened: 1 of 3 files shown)


holger-dengler:
New version with another implementation for the stack pointer initialization.

src/fips.c
```diff
@@ -1085,7 +1085,7 @@ ecdh_kat()
 	ica_adapter_handle_t ah;
 	const struct ecdh_kat_tv *tv;
 	unsigned char shared_secret[MAX_ECC_PRIV_SIZE];
-	ICA_EC_KEY *eckey_A, *eckey_B;
+	ICA_EC_KEY *eckey_A = NULL, *eckey_B = NULL;
```
ifranzki, line 1088: You must also set eckey_A and eckey_B to NULL after freeing it in lines 1134/1135.
Otherwise if it fails in the 2. loop iteration, eckey_B is still non-NULL, and gets freed twice.

holger-dengler: Hm, looks like the other solution was the better one...

holger-dengler: I see 2 options: initialize the pointers at the beginning of the for loop (instead of at the beginning of the function) or switch back to my first implementation. I personally prefer the latter one.

jschmidb: Hm, sorry, yes, maybe your first solution was indeed better ...

holger-dengler: @jschmidb its up to you to decide.

jschmidb: In this case it would not be necessary to initialize the keys to NULL, but I would do this anyway

holger-dengler: ok, I'll go back to the first implementation.

holger-dengler:
re-push of the first version. Sorry for the noise.

ifranzki (approved):

jschmidb (approved):

PR #92, opened 2022-09-02
Fix testcases failures in the CI

ifranzki:
Set openssl3_initialized = 1 before running the FIPS self tests.

Fixes: eee292c07bedb33c55e08bda95b9db9449525d6b

### DIFF
--- a/src/init.c
+++ b/src/init.c
@@ -130,6 +130,10 @@ void __attribute__ ((constructor)) icainit(void)
 	fips_init(); /* before powerup tests and initialize functionlist */
 #endif
 
+#if OPENSSL_VERSION_PREREQ(3, 0)
+	openssl3_initialized = 1;
+#endif
+
 	rng_init();
 
 	s390_prng_init();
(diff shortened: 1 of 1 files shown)


ifranzki:
I saw `lib ctx null or openssl 3 not initialized` when executing the tests.
This comes from macro BEGIN_OPENSSL_LIBCTX and causes it to return EFAULT.

This caused the FIPS selftests to fail, syslog showed  Libica `ECDSA test failed` and `Libica ECDH test failed`, and thus libica ends up in error state.

ifranzki:
I guess the fix (https://github.com/opencryptoki/libica/commit/084d8188bffa74d74f014c0d3f3dabf4bc23887b) from @holger-dengler fixed the segfault in the error path, but the tests did still fail, because we still run into the error case.

holger-dengler:
Nice!

Just a nit: I personally prefer the linux-way to refer to other commits in a commit message 
`eee292c07bed ("FIPS 140-3: change order of lib initialization and self-tests")`

ifranzki:
Yes, will change this. Actually its not only caused by this commit, but also by https://github.com/opencryptoki/libica/commit/f2db11f2236e1c1cc2e5fd937801fe50505d2233 which added the self tests for ECDSA and ECDH, those are now producing the error. So maybe I should not add a fixes tag at all ?

Also realized that the commit title is maybe not the really best....doesn't really tell what was changed.

ifranzki:
Commit title and message changed, no code change.

holger-dengler:
You're right, it looks like the problem occurred as side effects of the other commit. I'm ok with the removal of the references.

holger-dengler (approved):
ACK

jschmidb (approved):
lgtm, thanks for fixing!

PR #93, opened 2022-09-07
Fix FIPS checking for Ed/X functions

ifranzki:
Don't allow Ed/X functions if the libica is build with FIPS enabled and the system runs in FIPS mode.


jschmidb (approved):
lgtm

PR #95, opened 2022-09-20
ECC: ec_key_check() uses OpenSSL, thus use own library context

ifranzki:
The ec_key_check() routine internally uses OpenSSL calls. In case libica is called from an OpenSSL provider, avoid a recursive loop by using libica's own OpenSSL library context with the default or fips provider loaded.


jschmidb (approved):
lgtm

PR #96, opened 2022-09-20
ECC: Fix maybe-uninitialized warning

ifranzki:


jschmidb (approved):
yep ... I also oversaw this. Should be good now.

PR #99, opened 2022-10-01
added more files for "make dist", so we can do a full "make clean ; make"

msmeissn:
added missing EXTRA_DIST files needed for "make clean; make"

Signed-off-by: Marcus Meissner <meissner@suse.de>


holger-dengler (changes requested):
Hi Macus,

thanks for your PR. Please see also my other comments. I would like to ask you for doing the changes. Please also add a proper subject line and Signed-off-by: to your commit message.

Unfortunately I also found another bug. In buildpath replacements in icastats_test.c causes some troubles in clean build environments. I'm currently working on a fix for that problem as well an will provide a fix tomorrow. I would appreciate comments on that.

cu, Holger

holger-dengler (approved):
Look good to me. Thanks a lot, Marcus.
@jschmidb pls pull/merge

PR #100, opened 2022-10-04
AES-XTS: Support multi part operations

ifranzki:
Function ica_aes_xts() does not support multi-part operations (although the parameter description for the tweak parameter indicates that it would support multi-patrt operations, which is not true).

Add a new function ica_aes_xts_ex() that additionally takes the IV as parameter. For the very first part, the tweak must be specified. Internally the initial IV is calculated (via PCC instruction). After a successful operation on the first part, the resulting IV is returned, and must be passed to subsequent calls to ica_aes_xts_ex() to continue the operation.

For intermediate parts, parameter tweak must be NULL. Then, the IV passed as parameter is used for the operation. The IV is updated on return.

The first and intermediate parts must be a multiple on the AES block size. Only the last part can use partial blocks. Nevertheless, the last part must also be at least one full block.

### DIFF
--- a/include/ica_api.h
+++ b/include/ica_api.h
@@ -2952,7 +2952,7 @@ unsigned int ica_aes_cmac_last(const unsigned char *message, unsigned long messa
 			       unsigned int direction);
 
 /**
- * Encrypt or decrypt data with an AES key using the XEX Tweakable Bloc Cipher
+ * Encrypt or decrypt data with an AES key using the XEX Tweakable Block Cipher
  * with Ciphertext Stealing (XTS) mode as described in NIST Special Publication
  * 800-38E and IEEE standard 1619-2007.
  *
@@ -2976,7 +2976,7 @@ unsigned int ica_aes_cmac_last(const unsigned char *message, unsigned long messa
  * @param key1
  * Pointer to a buffer containing a valid AES key. key1 is used for the actual
  * encryption of the message buffer combined some vector computed from the
- * tweek value (Key1 in IEEE Std 1619-2007).
+ * tweak value (Key1 in IEEE Std 1619-2007).
  * @param key2
  * Pointer to a buffer containing a valid AES key key2 is used to encrypt the
  * tweak (Key2 in IEEE Std 1619-2007).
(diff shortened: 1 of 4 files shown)


include/ica_api.h
```diff
@@ -3005,6 +3002,71 @@ unsigned int ica_aes_xts(const unsigned char *in_data, unsigned char *out_data,
 			 unsigned int key_length, unsigned char *tweak,
 			 unsigned int direction);
 
+/**
+ * Encrypt or decrypt data with an AES key using the XEX Tweakable Bloc Cipher
```
jschmidb, line 3006: Block Cipher

include/ica_api.h
```diff
@@ -3005,6 +3002,71 @@ unsigned int ica_aes_xts(const unsigned char *in_data, unsigned char *out_data,
...
+ * Length in bytes of the message to be en/decrypted, which resides at the
+ * beginning of in_data. The minimal value of data_length is cipher block size
+ * (i.e. a multiple of 16 for AES).
+ * For multi-part operations, data_length must be a multiple of the cipher
+ * block size, unless for the final part. For the final part, the data_length
+ * must be at least one full cipher block.
+ * @param key1
+ * Pointer to a buffer containing a valid AES key. key1 is used for the actual
+ * encryption of the message buffer combined some vector computed from the
+ * tweek value (Key1 in IEEE Std 1619-2007).
```
jschmidb, line 3035: tweak

include/ica_api.h
```diff
@@ -3005,6 +3002,71 @@ unsigned int ica_aes_xts(const unsigned char *in_data, unsigned char *out_data,
...
+ * @param key_length
+ * The length in bytes of the AES key. For XTS supported AES key sizes are 16
+ * and 32 for AES-128 and AES-256 respectively.
+ * @param tweak
+ * Pointer to a valid 16 byte tweak value (as in IEEE Std 1619-2007).
+ * For multi-part operations the tweak must only be specified for the initial
+ * part. For subsequent parts, the tweak parameter must be NULL.
+ * @param iv
+ * Pointer to the initialization vector to be used for multi-part operations.
+ * If the tweak parameter is NULL, then the operations uses the initialization
```
jschmidb, line 3048: operation uses (or operations use?)

include/ica_api.h
ifranzki: fixed. Was already wrong in the existing comment.....

include/ica_api.h
ifranzki: fixed. Was already wrong in the existing comment.....

include/ica_api.h
ifranzki: fixed

jschmidb (approved):
Thanks for fixing!

PR #103, opened 2022-11-08
Tooling cleanup and c99 fixes

holger-dengler:
This RP covers a few tooling cleanups and fixes for c99-findings.

The first patch reworks the ignore patterns for git (`.gitignore`). With this commit, generated files are no longer displayed in `git status` after a build. This makes it much easier for a developer to keep track of changed files.

The second patch adds the files to the related clean targets, which are generated by coverage builds. This allows to reset to a clean source tree after coverage builds. This is also an improvement for developers and maybe also for package maintainers.

The third patch fixes c99-related warnings. It does not contain the replacement of the `asm` keyword (see notes to the next commit).

The last patch in this series enforce c99-related compiler checks for library and test code. This commit uses currently the standard option `-std=gnu99`, which is AFAIK gcc-specific.

Note: please treat the last patch as a proposal. We should discuss, if we stay with the gnu-specific option or switch to an ISO-style checking with the option `-std=c99`. If this is the case, there is another aspect we should discuss.

With the ISO-conform option `-std=c99`, we also have to care about the asm-keyword in the code. With this ISO-c99 option, `asm` is unknown, so the compiler will exit with an error. There are two possible options to fix the unknown keyword:
- add the option `-fasm`
- replace the keyword `asm` with `__asm__` all over the code.

**Summary of open questions:**
- Should we go with `-std=gnu99` or `-std=c99`?
- Should we add the option `-fasm` or should we replace the asm-keyword? (olny if the answer to the first question is c99)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,12 @@
+v4.2.0
+   [FEATURE] Display build info via icainfo -v 
+   [FEATURE] New API function ica_get_build_version()
+   [FEATURE] Display fips indication via icainfo -f
+   [FEATURE] New API function ica_get_fips_indicator()
+   [FEATURE] New API function ica_aes_gcm_initialize_fips()
+   [FEATURE] New API function ica_aes_gcm_kma_get_iv()
+   [FEATURE] New API function ica_get_msa_level()
+   [PATCH] icainfo: check for malloc error when getting functionlist
 v4.1.1
  - [PATCH] Fix aes-xts multi-part operations
    [PATCH] Fix make dist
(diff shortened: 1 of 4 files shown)


holger-dengler:
Correction: the option `-std=gnu99` is not gcc specific, clang also accepts this option. I would propose to continue with `std=gnu99`.

jschmidb:
There's a little typo in commit msg:

> build: remove coverage files in clean target
> The coverage build (<- build, not builds?) generates files, which should be removed during a
> clean target. Add the coverage patterns to the CLEANFILES variables.

jschmidb (approved):
Thanks for fixing Holger!

vkanjur (approved):

ifranzki (approved):
lgtm

holger-dengler:
> There's a little typo in commit msg:

fixed in the current version (force-push)

PR #104, opened 2022-11-28
Additional changes for fips 140-3 compliance

jschmidb:
This pr provides additional changes necessary for libica fips 140-3 compliance.


include/ica_api.h
```diff
@@ -3317,6 +3317,12 @@ kma_ctx* ica_aes_gcm_kma_ctx_new();
  * @param iv
  * Pointer to a readable buffer of size greater than or equal to iv_length
  * bytes, that contains an initialization vector of size iv_length.
+ * When running in fips mode, the pointer may alternatively be NULL, in which
```
ifranzki, line 3347: Well, this sentence is not 100% true:
_**When running in fips mode**, the pointer may alternatively be NULL_
Actually, the pointer can also be NULL when not in FIPS mode.....
Maybe rephrase this that for FIPS approved operation this pointer MUST be NULL or similar.

include/ica_api.h
```diff
@@ -3458,6 +3464,32 @@ ICA_EXPORT
 int ica_aes_gcm_kma_verify_tag(const unsigned char* known_tag, unsigned int tag_length,
 		const kma_ctx* ctx);
 
+/**
+ * Obtain the iv from the given context. This function is mainly intended to
+ * allow applications to query an internally created iv when running in fips
```
ifranzki, line 3501: also here, one can ALWAYS obtain the IV, not only when in FIPS mode. Maybe rephrase that the IV can be obtained when it was not specified during ica_aes_gcm_kma_init()

src/ica_api.c
```diff
@@ -4017,14 +4037,38 @@ int ica_aes_gcm_kma_verify_tag(const unsigned char* known_tag, unsigned int tag_
...
+	if (ctx == NULL)
+		return EINVAL;
+
+	if (iv == NULL) {
+		*iv_length = ctx->iv_length;
+		return 0;
+	}
+
+	if (*iv_length < ctx->iv_length)
+		return ENOMEM;
```
ifranzki, line 4142: ENOMEM? Shouldn't this be EINVAL? ENOMEM is only when allocating memory failed as far as I know.

src/ica_api.c
```diff
@@ -3711,6 +3711,30 @@ unsigned int ica_aes_gcm_initialize(const unsigned char *iv,
...
+#ifdef ICA_FIPS
+	if (fips >> 1)
+		return EACCES;
+	if (fips & ICA_FIPS_MODE) {
+		if (iv_length < GCM_RECOMMENDED_IV_LENGTH)
+			return EPERM;
+	}
+#endif
+
+	rc = RAND_bytes(iv, iv_length);
```
ifranzki, line 3770: Check for iv being NULL and return EINVAL if so? Or does RAND_bytes do this already? But better not rely on argument checking of an external library function.

src/s390_crypto.c
```diff
@@ -756,3 +760,32 @@ int s390_get_functionlist(libica_func_list_element *pmech_list,
...
+		return EINVAL;
+	}
+
+	if (!indicator_list) {
+		*indicator_list_len = sizeof(icaList) / sizeof(libica_fips_indicator_element);
+		return 0;
+	}
+
+	if (*indicator_list_len < (sizeof(icaList) / sizeof(libica_fips_indicator_element))) {
+		return ENOMEM;
```
ifranzki, line 780: ENOMEM? I would use EINVAL here. ENOMEM is for allocating memory only.

src/s390_ecc.c
```diff
@@ -396,10 +396,11 @@ static short get_default_domain(void)
 		return domain;
```
ifranzki, line 397: Lines 391 and 392: Isn't this dead code? Yit define variable domain in line 387 and initialize it to -1. 
The if in line 391 will never be true......

include/ica_api.h
jschmidb: Good catch: yes, I initially had this only for fips, but then decided to allow an internal iv also in non-fips mode. This gives applications more flexibility. But I didn't update the comment. Will do ...

include/ica_api.h
jschmidb: Yep, same as above, will change the text

src/ica_api.c
jschmidb: Hmm .... well, the ENOMEM here shall indicate that the application did not provide a "big-enough" buffer to receive the internally created iv. Which is indeed kind of a "space" or "memory" problem. I can change this to EINVAL if this is the agreed rc for this kind of error.

src/s390_crypto.c
jschmidb: ok, same discussion as above ... if I change it above, then also here.

src/s390_ecc.c
jschmidb: Yes, definitely. It's old code, but good to fix, maybe in a separate commit.

src/ica_api.c
jschmidb: Update: just had a look at the libzpc: there we use an own rc ZPC_ERROR_SMALLOUTBUF to indicate exactly this case where the application provides not enough space. So no conflict there.
When looking at the definition of ENOMEM:
`#define ENOMEM		12	/* Not enough space */`
it just says "not enough space". Hmmm ...

ifranzki: Well https://man7.org/linux/man-pages/man3/errno.3.html says "ENOMEM Not enough space/cannot allocate memory (POSIX.1-2001)."
https://www.gnu.org/software/libc/manual/html_node/Error-Codes.html says "ENOMEM
“Cannot allocate memory.” The system cannot allocate more virtual memory because its capacity is full."

jschmidb: hm, ok, then I will change to EINVAL at both places ...

include/ica_api.h
```diff
@@ -3270,6 +3276,12 @@ unsigned int ica_aes_gcm_initialize(const unsigned char *iv,
 					unsigned char *icb, unsigned char *ucb,
 					unsigned char *subkey, unsigned int direction);
 
+ICA_EXPORT
+unsigned int ica_aes_gcm_initialize_fips(unsigned char *iv,
+		unsigned int iv_length, unsigned char *key, unsigned int key_length,
```
ifranzki, line 3292: Indention of the parameter lines seems to be wrong

src/ica_api.c
```diff
@@ -3711,6 +3711,30 @@ unsigned int ica_aes_gcm_initialize(const unsigned char *iv,
 #endif /* NO_CPACF */
 }
 
+unsigned int ica_aes_gcm_initialize_fips(unsigned char *iv,
+		unsigned int iv_length, unsigned char *key, unsigned int key_length,
```
ifranzki, line 3755: indention of the parameter lines seems to be wrong

src/ica_api.c
```diff
@@ -3711,6 +3711,30 @@ unsigned int ica_aes_gcm_initialize(const unsigned char *iv,
...
+			return EPERM;
+	}
+#endif
+
+	rc = RAND_bytes(iv, iv_length);
+	if (rc != 1)
+		return EIO;
+
+	return ica_aes_gcm_initialize(iv, iv_length, key, key_length,
+						icb, ucb, subkey, direction);
```
ifranzki: indention seems to be wrong

src/ica_api.c
```diff
@@ -3711,6 +3711,30 @@ unsigned int ica_aes_gcm_initialize(const unsigned char *iv,
...
+}
+
 unsigned int ica_aes_gcm_intermediate(unsigned char *plaintext,
 			 unsigned long plaintext_length,
```
ifranzki, line 3779: indention also seems to be wrong here

src/fips.c
```diff
@@ -1227,7 +1227,8 @@ ecdsa_kat(void)
 		if (rc)
 			goto _err_;
 		/* adapter handle not needed here, just CPACF */
-		rc = ica_ecdsa_sign_ex(0, eckey, tv->hash, tv->hashlen, sigbuf, tv->siglen, tv->k);
+		rc = ica_ecdsa_sign_ex_internal(0, eckey, tv->hash, tv->hashlen,
+								sigbuf, tv->siglen, tv->k);
```
ifranzki, line 1231: indention

src/ica_api.c
```diff
@@ -1541,15 +1541,7 @@ int ica_ecdh_derive_secret(ica_adapter_handle_t adapter_handle,
...
-		const ICA_EC_KEY *privkey, const unsigned char *hash, unsigned int hash_length,
-		unsigned char *signature, unsigned int signature_length)
-{
-	return ica_ecdsa_sign_ex(adapter_handle, privkey, hash, hash_length,
-			signature, signature_length, NULL);
-}
-
-int ica_ecdsa_sign_ex(ica_adapter_handle_t adapter_handle,
+int ica_ecdsa_sign_ex_internal(ica_adapter_handle_t adapter_handle,
 		const ICA_EC_KEY *privkey, const unsigned char *hash, unsigned int hash_length,
```
ifranzki, line 1545: indention seems to be wrong here after renaming the function

src/ica_api.c
```diff
@@ -1619,6 +1611,28 @@ int ica_ecdsa_sign_ex(ica_adapter_handle_t adapter_handle,
 	return rc;
 }
 
+int ica_ecdsa_sign_ex(ica_adapter_handle_t adapter_handle,
+		const ICA_EC_KEY *privkey, const unsigned char *hash, unsigned int hash_length,
```
ifranzki, line 1615: indention

src/ica_api.c
```diff
@@ -1619,6 +1611,28 @@ int ica_ecdsa_sign_ex(ica_adapter_handle_t adapter_handle,
...
+		unsigned char *signature, unsigned int signature_length,
+		const unsigned char *k)
+{
+#ifdef ICA_FIPS
+	if (k != NULL && (fips & ICA_FIPS_MODE))
+		return EPERM;
+#endif
+
+	return ica_ecdsa_sign_ex_internal(adapter_handle, privkey, hash, hash_length,
+			signature, signature_length, k);
```
ifranzki, line 1625: indention

src/ica_api.c
```diff
@@ -1619,6 +1611,28 @@ int ica_ecdsa_sign_ex(ica_adapter_handle_t adapter_handle,
...
+}
+
+int ica_ecdsa_sign(ica_adapter_handle_t adapter_handle,
+		const ICA_EC_KEY *privkey, const unsigned char *hash, unsigned int hash_length,
```
ifranzki, line 1629: indention

src/ica_api.c
```diff
@@ -1619,6 +1611,28 @@ int ica_ecdsa_sign_ex(ica_adapter_handle_t adapter_handle,
...
+		unsigned char *signature, unsigned int signature_length)
+{
+	return ica_ecdsa_sign_ex_internal(adapter_handle, privkey, hash, hash_length,
+			signature, signature_length, NULL);
```
ifranzki, line 1633: indention

src/include/fips.h
```diff
@@ -34,6 +34,14 @@ void fips_init(void);
  */
 void fips_powerup_tests(void);
 
+/*
+ * Create deterministic ECDSA signatures in self-tests.
+ */
+int ica_ecdsa_sign_ex_internal(ica_adapter_handle_t adapter_handle,
+		const ICA_EC_KEY *privkey, const unsigned char *hash, unsigned int hash_length,
```
ifranzki, line 41: indention

include/ica_api.h
```diff
@@ -3829,6 +3837,40 @@ int ica_fips_status(void);
...
+ *    EINVAL if at least one invalid parameter is given
+ *    ENOMEM if the application provided buffer size (given in @fips_list_len)
+ *    is too small for the list.
+ *
+ * A typical usage scenario would be that a fips-aware application checks if
+ * a given mechanism is fips approved, before using it.
+ */
+ICA_EXPORT
+unsigned int ica_get_fips_indicator(libica_fips_indicator_element *fips_list,
+					unsigned int *fips_list_len);
```
ifranzki, line 3882: indention

src/ica_api.c
```diff
@@ -4430,4 +4430,10 @@ ica_fips_powerup_tests(void)
 	fips_powerup_tests();
 }
 
+unsigned int ica_get_fips_indicator(libica_fips_indicator_element *fips_list,
+					unsigned int *fips_list_len)
```
ifranzki, line 4487: indention

src/include/fips.h
```diff
@@ -50,10 +50,17 @@ static const int FIPS_BLACKLIST[] = {DES_ECB, DES_CBC, DES_CBC_CS, DES_OFB,
...
+    X448_KEYGEN, X448_DERIVE, RSA_ME, RSA_CRT, SHA512_DRNG };
 static const size_t FIPS_BLACKLIST_LEN = sizeof(FIPS_BLACKLIST)
     / sizeof(FIPS_BLACKLIST[0]);
 
+/*
+ * FIPS service indicator: List of tolerated but non-approved algorithms.
+ */
+static const int FIPS_OVERRIDE_LIST[] = { RSA_ME, RSA_CRT, SHA512_DRNG };
+static const size_t FIPS_OVERRIDE_LIST_LEN = sizeof(FIPS_OVERRIDE_LIST)
+	/ sizeof(FIPS_OVERRIDE_LIST[0]);
```
ifranzki, line 84: I would place the / on the first line, but thats just my personal opinion.....

src/include/s390_crypto.h
```diff
@@ -25,6 +25,11 @@
 
 #include <openssl/opensslv.h>
 
+#ifdef ICA_FIPS
+int s390_get_fips_indicator(libica_fips_indicator_element *fips_list,
+				unsigned int *fips_list_len);
```
ifranzki, line 30: indention

src/s390_crypto.c
```diff
@@ -674,7 +677,8 @@ int s390_get_functionlist(libica_func_list_element *pmech_list,
 	 *   approved.
 	 * - We are in an error state.
 	 * */
-	if (((fips & ICA_FIPS_MODE) && !fips_approved(pmech_list[x].mech_mode_id))
+	if (((fips & ICA_FIPS_MODE) && !fips_approved(pmech_list[x].mech_mode_id &&
+		!fips_override(pmech_list[x].mech_mode_id)))
```
ifranzki, line 681: indention?

src/s390_crypto.c
```diff
@@ -756,3 +760,32 @@ int s390_get_functionlist(libica_func_list_element *pmech_list,
   }
   return 0;
 }
+
+#ifdef ICA_FIPS
+int s390_get_fips_indicator(libica_fips_indicator_element *indicator_list,
+						unsigned int *indicator_list_len)
```
ifranzki, line 766: indention

src/s390_crypto.c
```diff
@@ -756,3 +760,32 @@ int s390_get_functionlist(libica_func_list_element *pmech_list,
...
+{
+	size_t i;
+
+	if (!indicator_list_len) {
```
ifranzki, line 770: no need for { ... } for a single line body

src/icainfo.c
```diff
@@ -452,6 +454,118 @@ static struct crypt_pair crypt_map[] = {
 	{NULL,0}
 };
 
+int get_index_in_fips_list(libica_fips_indicator_element *fips_list,
+		unsigned int fips_len, unsigned int algo_id)
```
ifranzki, line 458: indention

src/icainfo.c
```diff
@@ -452,6 +454,118 @@ static struct crypt_pair crypt_map[] = {
...
+	for (i = 0; i < fips_len; i++) {
+		if (fips_list[i].mech_mode_id == algo_id)
+			return i;
+	}
+
+	return -1;
+}
+
+int get_index_in_mech_list(libica_func_list_element *mech_list,
+		unsigned int mech_len, unsigned int algo_id)
```
ifranzki, line 471: indention

src/icainfo.c
```diff
@@ -452,6 +454,118 @@ static struct crypt_pair crypt_map[] = {
...
+		goto done;
+	}
+
+	/* Get fips indicator list */
+	if (ica_get_fips_indicator(NULL, &fips_len) != 0){
+		perror("get_fips_indicator: ");
+		goto done;
+	}
+
+//	if (fips_len != mech_len) {
```
ifranzki, line 521: remove comment

src/ica_api.c
```diff
@@ -3608,13 +3608,13 @@ unsigned int ica_aes_ccm(unsigned char *payload, unsigned long payload_length,
...
 
-unsigned int ica_aes_gcm(unsigned char *plaintext, unsigned long plaintext_length,
-			 unsigned char *ciphertext,
-			 const unsigned char *iv, unsigned int iv_length,
-			 const unsigned char *aad, unsigned long aad_length,
-			 unsigned char *tag, unsigned int tag_length,
-			 unsigned char *key, unsigned int key_length,
-			 unsigned int direction)
+unsigned int ica_aes_gcm_internal(unsigned char *plaintext,
+			unsigned long plaintext_length, unsigned char *ciphertext,
```
ifranzki, line 3612: indention

src/ica_api.c
```diff
@@ -3686,6 +3686,24 @@ unsigned int ica_aes_gcm(unsigned char *plaintext, unsigned long plaintext_lengt
 #endif /* NO_CPACF */
 }
 
+unsigned int ica_aes_gcm(unsigned char *plaintext,
+			unsigned long plaintext_length, unsigned char *ciphertext,
```
ifranzki, line 3690: indention is off by 1

src/ica_api.c
```diff
@@ -3686,6 +3686,24 @@ unsigned int ica_aes_gcm(unsigned char *plaintext, unsigned long plaintext_lengt
...
+			unsigned char *key, unsigned int key_length,
+			unsigned int direction)
+{
+#ifdef ICA_FIPS
+	if (fips & ICA_FIPS_MODE)
+		return EPERM;
+#endif /* ICA_FIPS */
+
+	return ica_aes_gcm_internal(plaintext, plaintext_length, ciphertext,
+			iv, iv_length, aad, aad_length, tag, tag_length,
```
ifranzki, line 3703: indention

src/include/fips.h
```diff
@@ -42,6 +42,18 @@ int ica_ecdsa_sign_ex_internal(ica_adapter_handle_t adapter_handle,
 		unsigned char *signature, unsigned int signature_length,
 		const unsigned char *k);
 
+/*
+ * Use ica_aes_gcm_internal for fips self-test KATs. External iv are
+ * allowed here.
+ */
+unsigned int ica_aes_gcm_internal(unsigned char *plaintext,
+			unsigned long plaintext_length, unsigned char *ciphertext,
```
ifranzki, line 50: indention

src/ica_api.c
```diff
@@ -3704,14 +3704,14 @@ unsigned int ica_aes_gcm(unsigned char *plaintext,
...
-unsigned int ica_aes_gcm_initialize(const unsigned char *iv,
-				    unsigned int iv_length,
-				    unsigned char *key,
-				    unsigned int key_length,
-				    unsigned char *icb,
-				    unsigned char *ucb,
-				    unsigned char *subkey,
-				    unsigned int direction)
+unsigned int ica_aes_gcm_initialize_internal(const unsigned char *iv,
+				unsigned int iv_length,
```
ifranzki, line 3708: indention

src/ica_api.c
```diff
@@ -3726,23 +3726,31 @@ unsigned int ica_aes_gcm_initialize(const unsigned char *iv,
...
-
 	function_code = aes_directed_fc(key_length, direction);
 
 	return s390_gcm_initialize(function_code, iv, iv_length,
 							   key, icb, ucb, subkey);
 #endif /* NO_CPACF */
 }
 
+unsigned int ica_aes_gcm_initialize(const unsigned char *iv,
+				unsigned int iv_length,
```
ifranzki, line 3737: indention

src/ica_api.c
```diff
@@ -3726,23 +3726,31 @@ unsigned int ica_aes_gcm_initialize(const unsigned char *iv,
...
+				unsigned char *subkey,
+				unsigned int direction)
+{
+#ifdef ICA_FIPS
+	if (direction == ENCRYPT && (fips & ICA_FIPS_MODE))
+		return EPERM;
+#endif /* ICA_FIPS */
+
+	return ica_aes_gcm_initialize_internal(iv, iv_length, key, key_length,
+									icb, ucb, subkey, direction);
```
ifranzki, line 3751: indention

src/ica_api.c
```diff
@@ -3763,8 +3771,8 @@ unsigned int ica_aes_gcm_initialize_fips(unsigned char *iv,
 	if (rc != 1)
 		return EIO;
 
-	return ica_aes_gcm_initialize(iv, iv_length, key, key_length,
-						icb, ucb, subkey, direction);
+	return ica_aes_gcm_initialize_internal(iv, iv_length, key, key_length,
+									icb, ucb, subkey, direction);
```
ifranzki, line 3775: indention?

src/include/fips.h
```diff
@@ -54,6 +54,16 @@ unsigned int ica_aes_gcm_internal(unsigned char *plaintext,
 			unsigned char *key, unsigned int key_length,
 			unsigned int direction);
 
+/*
+ * Use ica_aes_gcm_initialize_internal for fips self-test KATs. External iv are
+ * allowed here.
+ */
+unsigned int ica_aes_gcm_initialize_internal(const unsigned char *iv,
+				unsigned int iv_length, unsigned char *key,
```
ifranzki, line 62: indention

include/ica_api.h
```diff
@@ -3378,6 +3382,11 @@ int ica_aes_gcm_kma_init(unsigned int direction,
 					const unsigned char *key, unsigned int key_length,
 					kma_ctx* ctx);
 
+ICA_EXPORT
+int ica_aes_gcm_kma_init_fips(unsigned int direction, unsigned int iv_length,
+					const unsigned char *key, unsigned int key_length,
```
ifranzki, line 3387: indention

src/ica_api.c
```diff
@@ -3989,6 +3982,38 @@ int ica_aes_gcm_kma_init(unsigned int direction,
 #endif /* NO_CPACF */
 }
 
+int ica_aes_gcm_kma_init(unsigned int direction,
+					const unsigned char *iv, unsigned int iv_length,
```
ifranzki, line 3986: indention

src/ica_api.c
```diff
@@ -3989,6 +3982,38 @@ int ica_aes_gcm_kma_init(unsigned int direction,
...
+					const unsigned char *key, unsigned int key_length,
+					kma_ctx* ctx)
+{
+#ifdef ICA_FIPS
+	if (direction == ICA_ENCRYPT && (fips & ICA_FIPS_MODE))
+		return EPERM;
+#endif /* ICA_FIPS */
+
+	return ica_aes_gcm_kma_init_internal(direction, iv, iv_length,
+									key, key_length, ctx);
```
ifranzki, line 3996: indention

src/ica_api.c
```diff
@@ -3989,6 +3982,38 @@ int ica_aes_gcm_kma_init(unsigned int direction,
...
+}
+
+int ica_aes_gcm_kma_init_fips(unsigned int direction, unsigned int iv_length,
+					const unsigned char *key, unsigned int key_length,
```
ifranzki, line 4000: indention

src/ica_api.c
```diff
@@ -3989,6 +3982,38 @@ int ica_aes_gcm_kma_init(unsigned int direction,
...
+		return EACCES;
+
+	if (fips & ICA_FIPS_MODE) {
+		if (iv_length < GCM_RECOMMENDED_IV_LENGTH)
+			return EPERM;
+	}
+#endif /* ICA_FIPS */
+
+	return ica_aes_gcm_kma_init_internal(direction, NULL, iv_length,
+									key, key_length, ctx);
```
ifranzki, line 4014: indention

test/aes_gcm_test.c
```diff
@@ -704,6 +732,197 @@ int test_gcm_kat_update_in_place(int iteration)
...
+	if (ica_fips_status() & ICA_FIPS_MODE) {
+		if (iv_length < GCM_RECOMMENDED_IV_LENGTH) {
+			V_(printf("Skipping test because iv_length %d not allowed in fips mode.\n",
+					iv_length));
+			return TEST_SKIP;
+		}
+
+		/* Try to use an external iv for encrypt in fips mode: expect EPERM */
+		rc = ica_aes_gcm_initialize(iv, iv_length, key, key_length,
+								icb, ucb, subkey, ICA_ENCRYPT);
```
ifranzki, line 784: indention

test/aes_gcm_test.c
```diff
@@ -704,6 +732,197 @@ int test_gcm_kat_update_in_place(int iteration)
...
+	VV_(printf("Test Parameters for iteration = %i\n", iteration));
+	VV_(printf("key length = %i, data length = %i, tag length = %i,"
+		"iv length = %i aad_length = %i\n", key_length, data_length,
+		t_length, iv_length, aad_length));
+
+#ifdef ICA_FIPS
+	if (ica_fips_status() & ICA_FIPS_MODE) {
+		if (iv_length < GCM_RECOMMENDED_IV_LENGTH) {
+			V_(printf("Skipping test because iv_length %d not allowed in fips mode.\n",
+					iv_length));
```
ifranzki, line 778: indention

test/aes_gcm_test.c
```diff
@@ -704,6 +732,197 @@ int test_gcm_kat_update_in_place(int iteration)
...
+#endif
+
+	/* For encryption use ica_aes_gcm_initialize_fips() to create an internal
+	 * fips compliant iv, which is returned in *output* variable iv, rather than
+	 * specifying an external iv via *input* variable iv in
+	 * ica_aes_gcm_initialize(). */
+	aad_length_tmp = aad_length;
+	memset(running_tag, 0, AES_BLOCK_SIZE);
+	rc = ica_aes_gcm_initialize_fips(iv, iv_length, key, key_length,
+								icb, ucb, subkey, ICA_ENCRYPT);
```
ifranzki, line 800: indention

test/aes_gcm_test.c
```diff
@@ -704,6 +732,197 @@ int test_gcm_kat_update_in_place(int iteration)
...
+		return TEST_FAIL;
+	}
+
+	/* As the iv will change through the intermediate encryption steps,
+	 * save it now, because we will need it later for decryption. */
+	memcpy(iv_save, iv, iv_length);
+
+	if (num_chunks == 0 && aad_length > 0) {
+		rc = ica_aes_gcm_intermediate(input_data, 0, encrypt,
+								ucb, aad, aad_length,
```
ifranzki, line 812: indention

test/aes_gcm_test.c
```diff
@@ -704,6 +732,197 @@ int test_gcm_kat_update_in_place(int iteration)
...
+								key, key_length, subkey, ICA_ENCRYPT);
+	}
+
+	offset = 0;
+	for (i = 0; i < num_chunks; i++) {
+		chunk_len = gcm_kats[iteration].chunks[i];
+		chunk_data = input_data + offset;
+
+		rc = ica_aes_gcm_intermediate(chunk_data, chunk_len, encrypt + offset,
+									ucb, aad, aad_length,
```
ifranzki, line 823: indention

test/aes_gcm_test.c
```diff
@@ -704,6 +732,197 @@ int test_gcm_kat_update_in_place(int iteration)
...
+									running_tag, AES_BLOCK_SIZE,
+									key, key_length, subkey, ICA_ENCRYPT);
+		/* clear aad_length after first run*/
+		aad_length = 0;
+		offset += chunk_len;
+	}
+	sum_A_len = aad_length_tmp;
+	sum_C_len = offset;
+	rc = ica_aes_gcm_last(icb, sum_A_len, sum_C_len, running_tag,
+						t, t_length, key, key_length, subkey, ICA_ENCRYPT);
```
ifranzki, line 833: indention

test/aes_gcm_test.c
```diff
@@ -704,6 +732,197 @@ int test_gcm_kat_update_in_place(int iteration)
...
+	}
+
+	/* Save final tag from encryption for later use at decryption. The
+	 * running_tag variable is re-used and changed while decrypting. */
+	memcpy(t_result, running_tag, t_length);
+
+	aad_length = aad_length_tmp;
+	memset(running_tag, 0, AES_BLOCK_SIZE);
+	rc = ica_aes_gcm_initialize(iv_save, iv_length, key, key_length,
+								icb, ucb, subkey, ICA_DECRYPT);
```
ifranzki, line 864: indention

test/aes_gcm_test.c
```diff
@@ -704,6 +732,197 @@ int test_gcm_kat_update_in_place(int iteration)
...
+
+	if (num_chunks == 0 && aad_length > 0) {
+		rc = ica_aes_gcm_intermediate(input_data, 0, encrypt,
+								ucb, aad, aad_length,
```
ifranzki, line 868: indention

test/aes_gcm_test.c
```diff
@@ -704,6 +732,197 @@ int test_gcm_kat_update_in_place(int iteration)
...
+								key, key_length, subkey, ICA_DECRYPT);
+	}
+
+	offset = 0;
+	for (i = 0; i < num_chunks; i++) {
+		chunk_len = gcm_kats[iteration].chunks[i];
+		chunk_data = encrypt + offset;
+
+		rc = ica_aes_gcm_intermediate(decrypt + offset, chunk_len, chunk_data,
+									ucb, aad, aad_length,
```
ifranzki, line 879: indention

test/aes_gcm_test.c
```diff
@@ -704,6 +732,197 @@ int test_gcm_kat_update_in_place(int iteration)
...
+									key, key_length, subkey, ICA_DECRYPT);
+
+		/* clear aad_length after first run*/
+		aad_length = 0;
+		offset += chunk_len;
+	}
+	sum_A_len = aad_length_tmp;
+	sum_C_len = offset;
+	rc = ica_aes_gcm_last(icb, sum_A_len, sum_C_len, running_tag,
+					t_result, t_length, key, key_length, subkey, ICA_DECRYPT);
```
ifranzki, line 890: indention

src/icainfo.c
jschmidb: Yep! Forgot to remove it. Will do.

src/s390_crypto.c
jschmidb: A general remark on indention: yes, I also see all these places where parms are not aligned in the same column. 
First observation: in Eclipse it looks different, because the tab setting is different.
Second: many function names are quite long, so there is not much space left before col 80 to place the parm lists.
What I in fact don't like is mixing of tabs and spaces. And with tabs alone you mostly cannot precisely place subsequent parm lines right below the first one. For better readability I'm then moving such parm lines to the left.
So after all I would leave all these indentations as they are, for now. Even as we have a mixture of tabs and spaces at many places in old code. But I consider to do a global rework on this at a later time.

src/include/fips.h
jschmidb: I agree ... I never liked the coding in lines 76 + 77, which is old code, but wanted to add the new code in the same way. OK, this is the trigger to change both places ;-)

jschmidb: Update: I would code it like:
```
static const size_t FIPS_OVERRIDE_LIST_LEN
	= sizeof(FIPS_OVERRIDE_LIST) / sizeof(FIPS_OVERRIDE_LIST[0]);
```
Looks good to me, but this is also personal preference ...

src/s390_crypto.c
ifranzki: fine with me. I just would try to stay consistent over all files. 
I don't see any specific codding style guidelines defined for libica. For Opencryptoki we have that: https://github.com/opencryptoki/opencryptoki/blob/master/doc/coding_style.md
It seems that libica is using the linux kernel style, i.e. using tabs (8 places): https://www.kernel.org/doc/html/v4.10/process/coding-style.html
Nevertheless, define a style you want and then stick to it.

src/include/fips.h
ifranzki: I would place the = at the end of the first line, but other than that it looks good.

src/ica_api.c
```diff
@@ -4177,6 +4177,32 @@ int ica_get_msa_level(void)
...
+int ica_get_hw_info(libica_hw_info *hw_info)
+{
+	char line[32];
+	char dummy[32];
+	FILE *fd;
+
+	if (hw_info == NULL)
+		return EINVAL;
+
+	if ((fd = fopen("/proc/cpuinfo", "r")) == NULL)
```
ifranzki, line 4189: Well, this function is nonsense in my opinion.  Why adding a function to libica for something that can also be done out of the box with a simple `cat /proc/cpuinfo` ?
This function does not really add anything.

jschmidb: You are completely right, but fips certification requires this info to be available programmatically through the library API.

ifranzki: Hmm... OpenSSL seems to be FIPS certified, right? So where in OpenSSL is this API that you can programmatically query such information?????

jschmidb: The difference between openssl and libica is (as seen by the fips certifiers), that openssl is considered as a software module, while libica is considered as a "hybrid module", consisting of a software part (our libica code) and a hardware/firmware part, which is CPACF. For this reason they want the library to provide version levels of both parts. 

I already asked CPACF development whether there is an unambiguous version indication. They told me there's none. So all we have is the MSA level, which is not unique, because there are multiple processor generations with the same (highest) MSA level. But together with the processor id (machine generation) it sufficient for the certification.

vkanjur (approved):

PR #105, opened 2023-02-01
bugfix: permission denied on shared memory segments

jschmidb:
A change to the Linux kernel in 4.19 for added security is changing the behavior when opening shared memory segments. The O_CREAT flag must not be used for existing segments.

### DIFF
--- a/src/icastats_shared.c
+++ b/src/icastats_shared.c
@@ -54,9 +54,10 @@ int stats_mmap(int user)
 	sprintf(shm_id, "icastats_%d",
 		user == -1 ? geteuid() : (uid_t)user);
 
-	stats_shm_handle = shm_open(shm_id,
-				    O_CREAT | O_RDWR,
-				    S_IRUSR | S_IWUSR);
+	stats_shm_handle = shm_open(shm_id, O_RDWR,  S_IRUSR | S_IWUSR);
+
+	if (stats_shm_handle == -1)
+		stats_shm_handle = shm_open(shm_id, O_CREAT | O_RDWR, S_IRUSR | S_IWUSR);
 
 	if (stats_shm_handle == -1)
 		return rc;


src/icastats_shared.c
```diff
@@ -27,6 +27,7 @@
 #include "icastats.h"
 #include "init.h"
 
+#define NOT_INITIALIZED (-1)
```
holger-dengler, line 30: I would not define a macro for this. The `-1` is just a normal return value of a syscall.

src/icastats_shared.c
```diff
@@ -54,9 +55,11 @@ int stats_mmap(int user)
 	sprintf(shm_id, "icastats_%d",
 		user == -1 ? geteuid() : (uid_t)user);
 
-	stats_shm_handle = shm_open(shm_id,
-				    O_CREAT | O_RDWR,
-				    S_IRUSR | S_IWUSR);
+	stats_shm_handle = shm_open(shm_id, O_RDWR,  S_IRUSR | S_IWUSR);
+
+	if (stats_shm_handle == NOT_INITIALIZED)
```
holger-dengler, line 60: Please use -1 instead of the macro, like in line 64.

holger-dengler (approved):
Thanks for figuring that out.

See my other comments and feel free to drop them, if you would stay with the current solution.

src/icastats_shared.c
jschmidb: Yes, I added the line because this define was used in the old code. But I agree, as we are using the define only here, we may indeed just check for -1.

jschmidb: Fixed.

src/icastats_shared.c
```diff
@@ -54,9 +54,11 @@ int stats_mmap(int user)
...
 		user == -1 ? geteuid() : (uid_t)user);
 
-	stats_shm_handle = shm_open(shm_id,
-				    O_CREAT | O_RDWR,
-				    S_IRUSR | S_IWUSR);
+	stats_shm_handle = shm_open(shm_id, O_RDWR,  S_IRUSR | S_IWUSR);
+
+	if (stats_shm_handle == -1)
+		stats_shm_handle = shm_open(shm_id, O_CREAT | O_RDWR,
+											S_IRUSR | S_IWUSR);
```
ifranzki, line 61: I would say there are too many tabs.  With a tab size of 8 (which fits to all other places) there should be only 5 tabs.

jschmidb: Yep ... saw it also. In eclipse it looked ok.
Fixed.

ifranzki: I think its still too many tabs....
Make sure you editor setting displays a tab as 8 characters !

jschmidb: Well, I have 4 chars per tab and don't want to change this. Removing more tabs would make it look bad in eclipse. I know that for kernel development, 8 chars per tab are a must. But we never followed this strictly in libica.

ifranzki: Well, the remaining libica code uses 8 chars per tab.

See a few lines above:

```
	stats_shm_handle = shm_open(shm_id,
				    O_CREAT | O_RDWR,
				    S_IRUSR | S_IWUSR);
```

This is correctly aligned with 8 chars per tab, but wrong for 4 chars per tab. 

You can set the chars per tab in eclipse specific for each project, so that it looks good for every style.

jschmidb: Everything in one line now. BTW: the old code had 6 tabs plus 4 spaces, which I really don't like.

ifranzki: much better now! Thanks.

ifranzki (approved):

PR #110, opened 2023-04-25
Updates for RSA key checking for p > q

ifranzki:


jschmidb:
lgtm. 
I guess the comment is for the maybe rare, but not impossible case, where a key comes from outside libica and is converted into the libica CRT form before using the libica API.

jschmidb:
there's a small typo in commit msg:
Add note about possible key modification if ica_rsa_crt()
I a CRT key with p < q    <- If a  ...

ifranzki:
> I guess the comment is for the maybe rare, but not impossible case, where a key comes from outside libica and is converted into the libica CRT form before using the libica API.

Right, but also for the case that libica generates an RSA key using OpenSSL 3.0, which does generate keys with p < q (not with OpenSSL 1.1.1, but starting with 3.0....). With the second commit of this PR this is of course 'corrected' during key generation.

ifranzki:
> there's a small typo in commit msg:

fixed

jschmidb (approved):

ifranzki:
Can this be merged now?

jschmidb:
ah yes ... done.

PR #112, opened 2023-06-19
Replace deprecated ioctl

jschmidb:
Replace deprecated ioctl Z90STAT_STATUS_MASK by ZCRYPT_STATUS_MASK 
as this legacy ioctl will be removed soon.

### DIFF
--- a/src/ica_api.c
+++ b/src/ica_api.c
@@ -345,12 +345,11 @@ static unsigned int check_message_part(unsigned int message_part)
 
 unsigned int ica_open_adapter(ica_adapter_handle_t *adapter_handle)
 {
-	char *name, status_mask[64];
+	char *name;
 
 	if (!adapter_handle)
 		return EINVAL;
 
-	*adapter_handle = DRIVER_NOT_LOADED;
 	name = getenv("LIBICA_CRYPT_DEVICE");
 	if (name)
 		*adapter_handle = open(name, O_RDWR);
@@ -360,12 +359,6 @@ unsigned int ica_open_adapter(ica_adapter_handle_t *adapter_handle)
 			*adapter_handle = open(DEFAULT2_CRYPT_DEVICE, O_RDWR);
 		if (*adapter_handle == -1)
 			*adapter_handle = open(DEFAULT3_CRYPT_DEVICE, O_RDWR);
-	}
-	if (*adapter_handle != -1) {
-		/* Test if character device is accessible. */
-		if (!ioctl(*adapter_handle, Z90STAT_STATUS_MASK, &status_mask)) {
-			return 0;
-		}
 	}
 
 	/*


src/ica_api.c
```diff
@@ -363,7 +363,7 @@ unsigned int ica_open_adapter(ica_adapter_handle_t *adapter_handle)
 	}
 	if (*adapter_handle != -1) {
 		/* Test if character device is accessible. */
-		if (!ioctl(*adapter_handle, Z90STAT_STATUS_MASK, &status_mask)) {
+		if (!ioctl(*adapter_handle, ZCRYPT_STATUS_MASK, &status_mask)) {
 			return 0;
 		}
 	}
```
hfreude, line 369: What about this instead:

diff --git a/src/ica_api.c b/src/ica_api.c
index c9ea1ad..74e25b9 100644
--- a/src/ica_api.c
+++ b/src/ica_api.c
@@ -345,7 +345,8 @@ static unsigned int check_message_part(unsigned int message_part)
 
 unsigned int ica_open_adapter(ica_adapter_handle_t *adapter_handle)
 {
-       char *name, status_mask[64];
+       struct zcrypt_device_matrix_ext status_mask;
+       char *name;
 
        if (!adapter_handle)
                return EINVAL;
@@ -363,7 +364,7 @@ unsigned int ica_open_adapter(ica_adapter_handle_t *adapter_handle)
        }
        if (*adapter_handle != -1) {
                /* Test if character device is accessible. */
-               if (!ioctl(*adapter_handle, Z90STAT_STATUS_MASK, &status_mask)) {
+               if (!ioctl(*adapter_handle, ZCRYPT_DEVICE_STATUS, &status_mask)) {
                        return 0;
                }
        }

hfreude: Or you use your code but replace the status_mask[256] with status_mask[MAX_ZDEV_CARDIDS_EXT].
But I see that the ZCRYPT_STATUS_MASK ioctl will become deprecated soon also.

src/ica_api.c
```diff
@@ -363,7 +363,7 @@ unsigned int ica_open_adapter(ica_adapter_handle_t *adapter_handle)
...
 	if (*adapter_handle != -1) {
 		/* Test if character device is accessible. */
-		if (!ioctl(*adapter_handle, Z90STAT_STATUS_MASK, &status_mask)) {
+		if (!ioctl(*adapter_handle, ZCRYPT_STATUS_MASK, &status_mask)) {
```
ifranzki, line 366: The only question would be what sense it really makes to perform this check here at all. 
First: Why should the handle not be accessible if it could be opened? 
Second: Even if it is 'usable' now, this does not tell if it is still usable the next time you perform a request. A sandbox may have closed the fil handle in the meantime. 
But I guess thats a more philosophic question....

hfreude: Yes, I had same thoughts. First I thought, Jörg forgot to change the code which evaluates the content of the array but after having a look onto this code I am wondering if we need this at all. Why not let a failure happen on the 'real' ioctl ZSECSENDCPRB ?
Another option would be to change this code to evaluate something usefull - for example count the nr of online CCA cards and based on this maybe do not even try to send down a ZSECSENDCPRB ...

jschmidb: Yes, using MAX_ZDEV_CARDIDS_EXT is a good idea. It was just 64 before, so I first kept it similar. I'm wondering about the comment below the ioctl:
```
	/*
	 * Do not fail if crypto device driver is not loaded and CPACF is not
	 * available as the software fallback will still work without an adapter
	 * handle.
	 */
```
At least for libica-cex this is not true, but also libica has sw fallbacks disabled by default since quite some time. Maybe we have to rework this with another commit.

src/ica_api.c
```diff
@@ -361,12 +361,6 @@ unsigned int ica_open_adapter(ica_adapter_handle_t *adapter_handle)
 		if (*adapter_handle == -1)
 			*adapter_handle = open(DEFAULT3_CRYPT_DEVICE, O_RDWR);
 	}
-	if (*adapter_handle != -1) {
```
ifranzki, line 364: While you are at it:
The statement `*adapter_handle = DRIVER_NOT_LOADED;` does not make much sense. All code paths below will set `*adapter_handle`, so that assignment is superfluous. 
Nevertheless, `*adapter_handle` can be left at -1 (if the very last open fails), and -1 happens to be `DRIVER_NOT_LOADED`....

ifranzki (approved):
looks good now.

PR #113, opened 2023-07-18
icastats: Fix summary option to display correct summary information

ifranzki:
The '--summary' option of icastats did not display correct statistics since the introduction of per key keysize counters with libica version 4.0.0.

To display the correct summary counters, the all-key-size-counter values of an algorithm that supports multiple key sizes must be calculated like it is done in get_stats_data(). Adjust get_stats_data() function and friends so that it now also can be called from get_stats_sum() and can optionally operate on a specified statistics segment (i.e. the one where the summary statistics have been calculated in), not just the global one.

### DIFF
--- a/src/icastats.c
+++ b/src/icastats.c
@@ -302,7 +302,7 @@ int main(int argc, char *argv[])
 				perror("malloc: ");
 				return EXIT_FAILURE;
 			}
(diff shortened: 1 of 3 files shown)


holger-dengler (approved):
Looks good to me.

PR #114, opened 2023-08-23
Ensure OpenSSL config is loaded before creating libica's own library context

ifranzki:
Ensure OpenSSL is initialized and the OpenSSL config is loaded BEFORE creating libica's library context. Otherwise the OpenSSL config is loaded later, which may cause that all configured providers (e.g. the IBMCA provider) are also loaded into the library context. When the IBMCA provider is loaded into libica's library context, then a recursion can happen, since the IBMCA provider calls libica itself. Thus, we need to make sure that only the default or fips provider is loaded in the library context.

This change is important for applications that do not itself initialize OpenSSL. If they do, then libica's OpenSSL initialization is a no-op.

### DIFF
--- a/src/init.c
+++ b/src/init.c
@@ -118,7 +118,15 @@ void __attribute__ ((constructor)) icainit(void)
 	 * OpenSSL >= 3.0:
 	 * Create a separate library context for libica's use of OpenSSL services
 	 * and explicitly load the 'default' or 'fips' provider for this context.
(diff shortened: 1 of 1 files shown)


chrisRund (approved):
Fix worked as expected in our environment.

ifranzki:
@jschmidb I guess this PR is ready to merge.

PR #116, opened 2023-10-10
fips: use openssl lib context in compute_file_hmac

jschmidb:
Before calling any openssl function, the openssl lib ctx must be initialized. This was missing in compute_file_hmac.


ifranzki:
Maybe a little correction on the commit description: The openssl lib ctx is already initialized at that point in time, but we need to USE it here. The BEGIN macro sets the libica lib ctx as the current one for the current thread, so that it gets used by subsequent EVP calls. The END amcro changes it back to the previous lib context.

So maybe the following description would be better:

`Before calling any openssl EVP function, libica's own openssl lib ctx must be made the current one. 
This was missing in compute_file_hmac.`

jschmidb:
ok, commit text changed.

ifranzki (approved):

sharkcz:
I have asked the test team to recheck their tests with an updated libica package.

sharkcz (approved):
the coide makes sense tome, also our testing passed in FIPS mode

PR #117, opened 2023-11-14
icastats: Ignore unknown user in summary report

ifranzki:
If icastats is used with option '--summary' and it encounters a statistics shared memory segment for a user that no longer exists, it fails with 'get_stats_sum: : Success'. This is because getpwuid() fails, but does not set errno, so that perror prints 'Success'.

Instead of failing, just ignore such shared memory segments for user that do no longer exist. This behavior is consistent with the behavior of 'icastats --all' which shows statistics for all users, and also ignores shared memory segments for user that do no longer exist.

### DIFF
--- a/src/icastats_shared.c
+++ b/src/icastats_shared.c
@@ -263,10 +263,8 @@ int get_stats_sum(stats_entry_t *sum)
 			int fd;
 			stats_entry_t *tmp;
 
-			if ((getpwuid(atoi(&direntp->d_name[9]))) == NULL) {
-				closedir(shmDir);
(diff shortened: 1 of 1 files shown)


ifranzki:
Found more similar bugs for the `--all` case. Please take a look again.

jschmidb (approved):
yep, this one must also be changed. But as far as I can see, we now have all places.

chrisRund (approved):
lgtm, saw fix working as expected in comparison to behaviour in version 4.2.2

PR #118, opened 2023-11-20
Travis: change bionic (18.04) to focal (20.04).

jschmidb:
While at it, same as for libzpc.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,9 @@
-dist: bionic
+os: linux
+dist: focal
 
 language: c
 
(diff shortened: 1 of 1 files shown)


ifranzki (approved):
While you are at it, you could also fix the following minor Travis build config warnings:
```
Build config validation 
root: missing os, using the default linux
root: key matrix is an alias for jobs, using jobs
```

Other than that: lgtm

ifranzki:
Yep, Travis build config warnings are gone now. 
Still approved.

PR #119, opened 2023-11-21
ECC: fix KDSA sign behavior

jschmidb:
If we are in fips mode, the k-value is created via openssl. This may cause a cc=2 (random number not invertible) from the KDSA instruction. In this case loop over KDSA with new random values. But don't loop over calls to KDSA with unchanged k-value in the non-fips case. This could lead to an endless loop.

### DIFF
--- a/src/s390_ecc.c
+++ b/src/s390_ecc.c
@@ -1754,9 +1754,15 @@ struct {				\
 					rc = s390_kdsa(fc, param.buff, NULL, 0) ? EIO : 0;
 				} while ((rc != 0) && (++counter < MAX_KDSA_RETRIES));
 			} else {
+				/* Also create a random k-value for the non-deterministic
+				 * case as this number will be hashed with an internally
+				 * CPACF-created number and therefore increases quality.
+				 * Same below for p384 and p521. */
+				RAND_bytes(param.P256.rand + off, sizeof(param.P256.rand) - off);
 				rc = s390_kdsa(fc, param.buff, NULL, 0) ? EIO : 0;
 			}
 #else
+			RAND_bytes(param.P256.rand + off, sizeof(param.P256.rand) - off);
 			rc = s390_kdsa(fc, param.buff, NULL, 0) ? EIO : 0;
 #endif
 		} else {
@@ -1805,9 +1811,11 @@ struct {				\
 					rc = s390_kdsa(fc, param.buff, NULL, 0) ? EIO : 0;
 				} while ((rc != 0) && (++counter < MAX_KDSA_RETRIES));
 			} else {
+				RAND_bytes(param.P384.rand + off, sizeof(param.P384.rand) - off);
 				rc = s390_kdsa(fc, param.buff, NULL, 0) ? EIO : 0;
 			}
 #else
+			RAND_bytes(param.P384.rand + off, sizeof(param.P384.rand) - off);
 			rc = s390_kdsa(fc, param.buff, NULL, 0) ? EIO : 0;
 #endif
 		} else {
@@ -1867,9 +1875,11 @@ struct {				\
 					rc = s390_kdsa(fc, param.buff, NULL, 0) ? EIO : 0;
 				} while ((rc != 0) && (++counter < MAX_KDSA_RETRIES));
 			} else {
+				RAND_bytes(param.P521.rand + off, sizeof(param.P521.rand) - off);
 				rc = s390_kdsa(fc, param.buff, NULL, 0) ? EIO : 0;
 			}
 #else
+			RAND_bytes(param.P521.rand + off, sizeof(param.P521.rand) - off);
 			rc = s390_kdsa(fc, param.buff, NULL, 0) ? EIO : 0;
 #endif
 		} else {
--- a/src/s390_ecc.c
+++ b/src/s390_ecc.c
@@ -42,6 +42,7 @@ extern OSSL_LIB_CTX *openssl_libctx;
 
 #define CPRBXSIZE (sizeof(struct CPRBX))
 #define PARMBSIZE (2048)
+#define MAX_KDSA_RETRIES         10000
 
 
 static int eckeygen_cpacf(ICA_EC_KEY *key);
@@ -1715,7 +1716,7 @@ struct {				\
 #undef DEF_PARAM
 
 	unsigned long fc;
-	size_t off;
+	size_t off, counter = 0;
 	int rc;
 
 	memset(&param, 0, sizeof(param));
@@ -1743,16 +1744,25 @@ struct {				\
 #ifdef ICA_FIPS
 			if (fips & ICA_FIPS_MODE) {
 				fc |= 0x80; /* deterministic signature */
-				RAND_bytes(param.P256.rand + off, sizeof(param.P256.rand) - off);
+				do {
+					/* If the random number is not invertible, s390_kdsa will
+					 * cause to loop. Same for p384 and p521 below. */
+					if (RAND_bytes(param.P256.rand + off, sizeof(param.P256.rand) - off) != 1) {
+						rc = EIO;
+						break;
+					}
+					rc = s390_kdsa(fc, param.buff, NULL, 0) ? EIO : 0;
+				} while ((rc != 0) && (++counter < MAX_KDSA_RETRIES));
+			} else {
+				rc = s390_kdsa(fc, param.buff, NULL, 0) ? EIO : 0;
 			}
+#else
+			rc = s390_kdsa(fc, param.buff, NULL, 0) ? EIO : 0;
 #endif
-			rc = s390_kdsa(fc, param.buff, NULL, 0);
 		} else {
 			fc |= 0x80; /* deterministic signature */
-			do {
-				memcpy(param.P256.rand + off, k, sizeof(param.P256.rand) - off);
-				rc = s390_kdsa(fc, param.buff, NULL, 0);
-			} while (rc);
+			memcpy(param.P256.rand + off, k, sizeof(param.P256.rand) - off);
+			rc = s390_kdsa(fc, param.buff, NULL, 0) ? EIO : 0;
 		}
 
 		memcpy(sig, param.P256.sig_r + off,
@@ -1787,16 +1797,23 @@ struct {				\
 #ifdef ICA_FIPS
 			if (fips & ICA_FIPS_MODE) {
 				fc |= 0x80; /* deterministic signature */
-				RAND_bytes(param.P384.rand + off, sizeof(param.P384.rand) - off);
+				do {
+					if (RAND_bytes(param.P384.rand + off, sizeof(param.P384.rand) - off) != 1) {
+						rc = EIO;
+						break;
+					}
+					rc = s390_kdsa(fc, param.buff, NULL, 0) ? EIO : 0;
+				} while ((rc != 0) && (++counter < MAX_KDSA_RETRIES));
+			} else {
+				rc = s390_kdsa(fc, param.buff, NULL, 0) ? EIO : 0;
 			}
+#else
+			rc = s390_kdsa(fc, param.buff, NULL, 0) ? EIO : 0;
 #endif
-			rc = s390_kdsa(fc, param.buff, NULL, 0);
 		} else {
 			fc |= 0x80; /* deterministic signature */
-			do {
-				memcpy(param.P384.rand + off, k, sizeof(param.P384.rand) - off);
-				rc = s390_kdsa(fc, param.buff, NULL, 0);
-			} while (rc);
+			memcpy(param.P384.rand + off, k, sizeof(param.P384.rand) - off);
+			rc = s390_kdsa(fc, param.buff, NULL, 0) ? EIO : 0;
 		}
 
 		memcpy(sig, param.P384.sig_r + off,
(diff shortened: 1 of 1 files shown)


src/s390_ecc.c
```diff
@@ -1743,16 +1743,17 @@ struct {				\
 #ifdef ICA_FIPS
 			if (fips & ICA_FIPS_MODE) {
 				fc |= 0x80; /* deterministic signature */
-				RAND_bytes(param.P256.rand + off, sizeof(param.P256.rand) - off);
+				do {
+					RAND_bytes(param.P256.rand + off, sizeof(param.P256.rand) - off);
+					rc = s390_kdsa(fc, param.buff, NULL, 0);
+				while (rc)
```
ifranzki, line 1749: You don't want to do below s390_kdsa() again if you exit out of this loop. 
I guess the below s390_kdsa() needs to be on the else case of the above if.

ifranzki: same at the other 2 places for the other curves of course.

jschmidb: yes, right ...

src/s390_ecc.c
```diff
@@ -1743,16 +1743,19 @@ struct {				\
...
-				RAND_bytes(param.P256.rand + off, sizeof(param.P256.rand) - off);
+				do {
+					RAND_bytes(param.P256.rand + off, sizeof(param.P256.rand) - off);
+					rc = s390_kdsa(fc, param.buff, NULL, 0);
+				} while (rc);
+			} else {
+				rc = s390_kdsa(fc, param.buff, NULL, 0);
 			}
 #endif
 			rc = s390_kdsa(fc, param.buff, NULL, 0);
```
ifranzki, line 1754: still not correct. you still execute s390_kdsa in line 1754 when you come out of the if/else block.

I guess the s390_kdsa in line 1754 must go into an #else .... #endif of the #ifdef ICA_FIPS block.

src/s390_ecc.c
```diff
@@ -1750,9 +1750,14 @@ struct {				\
 					rc = s390_kdsa(fc, param.buff, NULL, 0);
 				} while (rc);
 			} else {
+				/* Also create a random k-value for the non-deterministic
+				 * case as this number will be XORed with an internally
```
ifranzki, line 1754: Not sure if it is XORed.... The POP states: "_specified random number is hashed with a CPU-generated random number_", so I would not mention XOR here.

The same comment would belong to line 1760.... but having it twice is certainly not really needed.

src/s390_ecc.c
```diff
@@ -1743,16 +1743,22 @@ struct {				\
 #ifdef ICA_FIPS
 			if (fips & ICA_FIPS_MODE) {
 				fc |= 0x80; /* deterministic signature */
-				RAND_bytes(param.P256.rand + off, sizeof(param.P256.rand) - off);
+				do {
+					/* A cc=2 occurs if the random number is not invertible.
+					 * Loop in this case. */
+					RAND_bytes(param.P256.rand + off, sizeof(param.P256.rand) - off);
+					rc = s390_kdsa(fc, param.buff, NULL, 0);
+				} while (rc);
```
ifranzki, line 1751: I guess we should only loop  for KDSA CC=2 here. If we also loop for other error cases (e.g. invalid private key) it will loop forever.

ifranzki: We might also add a repeat counter, and stop repeating after a certain number of repeats. Just in case the random number generator is faulty and always generates the same non-invertible nonce.

src/s390_ecc.c
```diff
@@ -1743,16 +1743,22 @@ struct {				\
 #ifdef ICA_FIPS
 			if (fips & ICA_FIPS_MODE) {
 				fc |= 0x80; /* deterministic signature */
-				RAND_bytes(param.P256.rand + off, sizeof(param.P256.rand) - off);
+				do {
+					/* A cc=2 occurs if the random number is not invertible.
+					 * Loop in this case. */
+					RAND_bytes(param.P256.rand + off, sizeof(param.P256.rand) - off);
```
ifranzki, line 1749: I guess we should check the error of RAND_bytes() here. For the FIPS case it is really essential that the random number generator did work, so we should error out if not.

For the other cases, RAND_bytes is just used to get a better random quality for the nonce and it does not really matter if it fails. KDSA will anyway generate a random nonce internally.

src/s390_ecc.c
jschmidb: ok, changed comment. And we had it even 3 times. I added a "same for p384 and p521" and removed 2 comments.

src/s390_ecc.c
jschmidb: Yes, a counter definitely makes sense.

ifranzki: this could then also cover the 'invalid private key' error case. If it ever happens that a wrong key gets that far into this routine, then it would only loop until the counter hits and fail that way. 

Still, we also could adjust the s390_kdsa function to return the CC to the caller, as suggested in a private email to you.

src/s390_ecc.c
ifranzki: also change the commit description please, you also mention XOR there.

src/s390_ecc.c
jschmidb: ok, done.

src/s390_ecc.c
jschmidb: counter is done, but the cc=2 handling is still in discussion

src/s390_ecc.c
```diff
@@ -1743,16 +1744,26 @@ struct {				\
...
+					 * Loop in this case. Same for p384 and p521 below. */
+					counter++;
+					if (RAND_bytes(param.P256.rand + off, sizeof(param.P256.rand) - off) != 1) {
+						rc = EIO;
+						break;
+					}
+					rc = s390_kdsa(fc, param.buff, NULL, 0);
+				} while (rc != 0 && counter < MAX_KDSA_RETRIES);
+			} else {
+				rc = s390_kdsa(fc, param.buff, NULL, 0);
```
ifranzki, line 1763: s390_kdsa returns 0 on success and 1 on error, and we return it as is from ecdsa_sign_cpacf. However, isn't ecdsa_sign_cpacf supposed to return an errno value in case of an error. 1 is EPERM (Operation not permitted).... is this what we want to return? Maybe we should change s390_kdsa to initialize rc to EIO instead? EINVAL could fit better (i.e. invalid deterministic nonce), but EINVAL means curve not supported as per caller  ecdsa_sign_hw....

src/s390_ecc.c
```diff
@@ -1830,19 +1849,25 @@ struct {				\
...
-				 * Use one byte less */
 				fc |= 0x80; /* deterministic signature */
-				RAND_bytes(param.P521.rand + off + 1, sizeof(param.P521.rand) - off - 1);
+				do {
+					counter++;
+					if (RAND_bytes(param.P521.rand + off, sizeof(param.P521.rand) - off) != 1) {
+						rc = EIO;
+						break;
+					}
+					rc = s390_kdsa(fc, param.buff, NULL, 0);
```
ifranzki, line 1867: In order to avoid many retries caused by nonces that are greater than the curve order 'n' (cause CC=2 with KDSA) we need to do something special here for P-521:

For P-521 the order 'n' is `0x01fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffa51868783bf2f966b7fcc0148f709a5d03bb5c9b8899c47aebb6fb71e91386409`, thus, its first bytes is 0x01, so the first seven bits are off, only the 8th bit is on, resulting in a bit length of 521 bits (no wonder for P-521, right?).
Any random nonce that has anything else than 0x01 or 0x00 in the very first byte will be larger than the order, and thus KDSA will fail with CC=2 and the loop takes a retry. 

Add the following line just before the KDSA call to set the first bits to zero (maybe with a proper comment):
`*(param.P521.rand + off) &= 0x01;`
That way the random nonce is at least not larger than 521 bits. It could still be numerically larger than the order, but then the loop will run a retry, which is OK.

For the P-256 and P-384 cases this special logic is NOT require, because the order of those curves start with 0xffffff..., so all 8 bits of the first byte are used, and thus no making is required.

src/s390_ecc.c
jschmidb: ok, done.

src/s390_ecc.c
```diff
@@ -1743,16 +1744,26 @@ struct {				\
 #ifdef ICA_FIPS
 			if (fips & ICA_FIPS_MODE) {
 				fc |= 0x80; /* deterministic signature */
-				RAND_bytes(param.P256.rand + off, sizeof(param.P256.rand) - off);
+				do {
+					/* A cc=2 occurs if the random number is not invertible.
+					 * Loop in this case. Same for p384 and p521 below. */
```
holger-dengler, line 1749: Should we mention `cc` here? The value is only visible in the inline-assembly function and is never returned to the caller.

src/include/s390_crypto.h
```diff
@@ -625,7 +625,7 @@ static inline int s390_kdsa(unsigned long func, void *param,
 	register unsigned long r1 asm("1") = (unsigned long)param;
 	register unsigned long r2 asm("2") = (unsigned long)src;
 	register unsigned long r3 asm("3") = (unsigned long)srclen;
-	unsigned long rc = 1;
+	unsigned long rc = 5;
```
ifranzki, line 628: Why not use EIO here (instead of hardcoded 5 ) ????

jschmidb: Because it does not compile ... errno.h is not yet included ... include it for this single place?

jschmidb: ... Or add a comment, like rc = 5; /* EIO */

holger-dengler: I just realized, that libica-api does not follow the best-practice for return values (rc/errno combination) like it is done in e.g. glibc. We should think about a change here for the next major release.

holger-dengler (commented):
Some minor comments. I think, the more interesting part is, if we should return the cc in the inline assembly function and do the errno mapping in the calling functions.

src/include/s390_crypto.h
holger-dengler: I would prefer to leave the rc = 1 unchanged and do the maooing in the calling function. e.g.
```
if (s390_kdsa(...)) {
    rc == EIO;
    ...
}
```

ifranzki: maybe `rc = s390_kdsa(....) ? EIO : 0;`  to keep the changes as small as possible. There are several places where kdsa is called.

jschmidb: When looking at this, we have 7 routines where s390_kdsa is called. In 6 of them the rc from s390_kdsa is immediately changed/set to a desired value, based on rc == 0 or != 0. So no change necessary here.
The remaining place is in fact ecdsa_sign_cpacf (several occurrences in this routine). Here I would do a rc = s390_kdsa(....) ? EIO : 0; as Ingo suggested. 
Let me know if you agree.

src/s390_ecc.c
```diff
@@ -1743,16 +1744,26 @@ struct {				\
...
+					counter++;
+					if (RAND_bytes(param.P256.rand + off, sizeof(param.P256.rand) - off) != 1) {
+						rc = EIO;
+						break;
+					}
+					rc = s390_kdsa(fc, param.buff, NULL, 0);
+				} while (rc != 0 && counter < MAX_KDSA_RETRIES);
```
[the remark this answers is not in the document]
jschmidb: Hm, I'm not so much into double parentheses, but ++counter in the while condition makes sense. OK, maybe it's in fact better readable with the ++counter ... I'll make the change, then let's have a look again ...

src/s390_ecc.c
jschmidb: ok, yes, I will change the comment.

src/s390_ecc.c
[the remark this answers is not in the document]
jschmidb: Done.

jschmidb:
> I just realized, that libica-api does not follow the best-practice for return values (rc/errno combination) like it is done in e.g. glibc. We should think about a change here for the next major release.

Added to my todo list ;-)

src/include/s390_crypto.h
holger-dengler: In `ecdsa_verify_cpacf()` we have also a conditional calling pattern, only the `rc` is changed to EFAULT. I agree to change it according to your proposal, but we should also agree upon the error value (EIO or EFAULT). In my opinion, EFAULT is not the right one (neither here, nor in `ecdsa_verify_cpacf()`), so I would vote for EIO.

BTW: Isn't changing the return value an API/ABI break?

src/s390_ecc.c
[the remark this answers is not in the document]
holger-dengler: in my opinion, it is easier to read with double parenthesis, but in the end, it's your decision.

src/include/s390_crypto.h
ifranzki: The verify case is different. Here `EFAULT` means that the signature is invalid (kind of an expected error case). 
`ecdsa_verify_cpacf()` is called by `ecdsa_verify_hw()` which is called by external API function `ica_ecdsa_verify()` and there `EFAULT` is documented as `signature invalid`. So this is part of the external API and can not be changed.

holger-dengler: from errno-base.h:
```
...
#define EIO              5      /* I/O error */
...
#define EFAULT          14      /* Bad address */
```

holger-dengler: Maybe `EINVAL`?

ifranzki: EINVAL is already used for unsupported curve.

jschmidb: With the recent push I fixed the rc = s390_kdsa(....) ? EIO : 0; for sign. No change for verify.

ifranzki (approved):
lgtm

holger-dengler (approved):
Looks good to me.

PR #120, opened 2023-12-11
Rework gcm parm checking

jschmidb:
This pull request does some rework on gcm parm checking for the old and new gcm API:
Some length constants were wrong, some checks were wrong, some checks were missing, and there were some typos in comments.

### DIFF
--- a/src/ica_api.c
+++ b/src/ica_api.c
@@ -258,24 +258,16 @@ static unsigned int check_gcm_parms(unsigned long text_length,
 				    const unsigned char *tag, unsigned int tag_length,
 				    unsigned int iv_length)
 {
-#ifdef __s390x__
 	/*
-	 * The following comparisions are alaways false on s390 targets
-	 * due to limited range of data type.
-	 */
-	if ((text_length > S390_GCM_MAX_TEXT_LENGTH) ||
-	    (aad_length > S390_GCM_MAX_AAD_LENGTH))
-		return EINVAL;
-#else
-	(void)text_length;	/* supporess unused param warning */
-	(void)aad_length;
-#endif
-	/*
-	 * The following check must be done but is commented out because
-	 * comparison is always false due to limited range of data type.
+	 * The following comparisons are always false due to limited
+	 * range of data types.
 	 *
-	 * if (iv_length   > S390_GCM_MAX_IV_LENGTH)
-	 *	return EINVAL;
+	 * if ((text_length > S390_GCM_MAX_TEXT_LENGTH) ||
+	 *     (aad_length > S390_GCM_MAX_AAD_LENGTH))
+	 *     return EINVAL;
+	 *
+	 * if (iv_length > S390_GCM_MAX_IV_LENGTH)
+	 *     return EINVAL;
 	 */
 
 	if (iv_length == 0)
--- a/src/include/s390_gcm.h
+++ b/src/include/s390_gcm.h
@@ -16,9 +16,23 @@
 
 #include "s390_ctr.h"
 
-#define S390_GCM_MAX_TEXT_LENGTH (0x0000000fffffffe0ul) /* (2^31)-32 */
-#define S390_GCM_MAX_AAD_LENGTH  (0x2000000000000000ul) /* (2^61)    */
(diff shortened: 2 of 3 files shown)


src/include/s390_gcm.h
```diff
@@ -681,8 +681,8 @@ static inline int s390_aes_gcm_simulate_kma_intermediate(const unsigned char *in
 	if (rc)
 		return rc;
 
-	ctx->total_aad_length += aad_length;
-	ctx->total_input_length += data_length;
+	ctx->total_aad_length += aad_length * 8;
```
ifranzki, line 698: So it seems that CPACF actually would support AAD and input text on bit boundary? 
However, the libica API only supports passing AAD and input text on byte boundary? 

I guess this is OK (and we would need an API break to add support for GCM on bit boundary anyway....).

jschmidb: I'm not sure: NIST says: 

> Although GCM is defined on bit strings, the bit lengths of the plaintext, the AAD, and the IV
> shall all be multiples of 8, so that these values are byte strings.

https://nvlpubs.nist.gov/nistpubs/legacy/sp/nistspecialpublication800-38d.pdf
(page 8)

ifranzki: OK, strange that CPACF then defines these fields as number of bits....

Also, doesn't this change mean that it should have calculated a wrong GCM output before? 
Don't we have known-answer-tests for GCM? They should have catched that, right? If we don't have them, maybe its time to introduce them now.

jschmidb: Yes, we have gcm known answer tests, but all with byte lengths ... not sure if there are NIST test vectors with bit lengths, as NIST itself says that the lengths should be byte lengths. In libica we enforce byte lengths anyway, but must multiply with 8 for the CPACG parm block.

jschmidb: Yes, good point! In fact, when calculating the final tag in s390_gcm_authenticate_last, the lengths are multiplied by 8, so it's correct to leave them as byte values in the ctx.

jschmidb: I removed the related commit.

src/ica_api.c
```diff
@@ -292,6 +292,16 @@ static unsigned int check_gcm_parms(unsigned long text_length,
 	return 0;
 }
 
+static unsigned int check_gcm_kma_parms(const kma_ctx* ctx, unsigned long data_length,
+									unsigned long aad_length)
+{
+	if (ctx->total_aad_length / 8 + aad_length > S390_GCM_MAX_AAD_LENGTH ||
```
ifranzki, line 298: is the `/ 8` still correct here? Isn't it first added to `ctx->total_aad_length` in full bytes, and only right before the instruction is called is multiplied by 8 to turn it into number of bits? If so, then I guess you should not divide by 8 here.

Same for `total_input_length` of course.

jschmidb: Hmm yes ... when using the "new" API, the lengths are already accumulated as bit values in routine s390_aes_gcm_kma:
```
	/* Actual lengths needed by KMA */
	ctx->total_aad_length += aad_length * 8;
	ctx->total_input_length += data_length * 8;

	/* Call KMA */
	rc = s390_kma(hw_fc, ctx,
			out_data, in_data, data_length,
			aad, aad_length);
```
The KMA instruction needs the values in bits. So we have a discrepancy between the old and new API concerning these two fields. In the old API they are byte values and converted to bits when calculating the final tag, but in the new API they need to be bit values for every KMA call. However, when thinking about this, only the final call to KMA would actually use the two values. 

So it might be possible to have byte values in the CPACF parm block until the last call. But I would leave it as is for now. 

At the end, the division by 8 in check_gcm_kma_parms would then be ok. And check_gcm_kma_parms is only called from the new API.

jschmidb: Correction: the discrepancy is also and mainly between the simulation path and KMA path in the new API. And yes, unfortunately in the simulation case we have byte values in the ctx while the KMA path has bit values. So the new check_gcm_kma_parms routine cannot be implemented in this way. Thanks for seeing this!
I'm removing the last commit.

ifranzki (approved):
lgtm

PR #121, opened 2024-01-05
GCM: new API function ica_allow_external_gcm_iv_in_fips_mode

jschmidb:
When running in fips mode, the GCM iv is created internally via an approved random source. Applications are not allowed to use an own, external iv. The new API function allows to override this behavior and allow an external GCM iv in fips mode. In this case the application is responsible for creating the iv in a compliant way.


ifranzki (approved):

ifranzki:
close/reopen to trigger travis.

PR #123, opened 2024-01-10
Initialize mechanism list before initializing OpenSSL in icainit

ifranzki:
In libica's library constructor OpenSSL is initialized since commit 05c4e36b1208193cd11fbc39347835c5a112c71b "Ensure OpenSSL config is loaded before creating libica's own library context". This will also cause OpenSSL to evaluate the OpenSSL config file and load providers configured there.

Even without this commit, it can happen when libica is in FIPS mode, that OpenSSL gets implicitly initialized by OpenSSL usage within the FIPS power on tests, which run as part of the library constructor.

If OpenSSL has been initialized by the application already, then the attempt to initialize OpenSSL again by libica is a no-operation. However, if the libica constructor is the very first to initialize OpenSSL this can cause problems. For applications that link against libica (via -lica), the libica library constructor is run before the applications main function, and thus it will most likely be the very first OpenSSL initialization attempt.

If the IBMCA provider is configured in the OpenSSL config file, then it is loaded and initialized during OpenSSL initialization triggered by the libica constructor. During IBMCA provider initialization, it loads libica and queries libica's mechanism list.

The problem is that this happens while we are still in the middle of the libica library constructor, and haven't fully initialized libica. Especially the mechanism list has not yet been set up, and thus the IBMCA provider will get a mechanism list where all algorithms are marked as being not supported. The IBMCA provider will thus not register any of its algorithms, and thus can not be used subsequentially by the application.

Unfortunately OpenSSL evaluates the config file only once, and also the IBMCA provider queries the libica mechanism list only once, so the situation will not change for the lifetime of the calling application.

To fix this, make sure that the mechanism list is initialized before the libica constructor initializes OpenSSL. That way, the IBMCA provider loaded due to the OpenSL initialization attempt from within the libica library constructor will get a valid mechanism list.

Since some mechanisms are dependent on the FIPS flag of libica, it must evaluate the system's FIPS state before it initializes OpenSSL, but finalize the FIPS setup (loading the FIPS provider, etc) only after OpenSSL has been initialized. For that the fips_init() function has been split into two parts.

### DIFF
--- a/src/fips.c
+++ b/src/fips.c
@@ -249,7 +249,7 @@ SHA3_KAT(512, 512);
 #endif
 
 void
-fips_init(void)
+fips_get_indicator(void)
 {
 	FILE *fd;
 	char fips_flag;
@@ -271,18 +271,22 @@ fips_init(void)
 		fips_flag = '1';
 
 	if (fips_flag - '0') {
-#if !OPENSSL_VERSION_PREREQ(3, 0)
 		/* Set libica into FIPS mode. */
 		fips |= ICA_FIPS_MODE;
+	}
+}
 
+void
+fips_init(void)
+{
+	if (fips & ICA_FIPS_MODE) {
+#if !OPENSSL_VERSION_PREREQ(3, 0)
 		/* Try to set OpenSSL into FIPS mode. If this is not possible,
 		 * all software fallbacks (including RSA key generation) will
 		 * be disabled. OpenSSL FIPS mode can be queried using the
 		 * FIPS_mode() function. */
 		FIPS_mode_set(1);
 #else
-		fips = 0;
-
 #ifndef NO_FIPS_CONFIG_LOAD
 		/* Allow to skip reading the openssl 3.x fips config. Tests showed
 		 * that this step must be skipped on RHEL9 systems. But on other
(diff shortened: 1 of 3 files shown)


src/fips.c
```diff
@@ -271,44 +271,46 @@ fips_init(void)
...
-
 #ifndef NO_FIPS_CONFIG_LOAD
 		/* Allow to skip reading the openssl 3.x fips config. Tests showed
 		 * that this step must be skipped on RHEL9 systems. But on other
 		 * systems, or with a locally built openssl, this step is necessary. */
 		if (!OSSL_LIB_CTX_load_config(openssl_libctx, LIBICA_FIPS_CONFIG)) {
 			syslog(LOG_ERR, "Libica failed to load openssl fips config %s\n",
 					LIBICA_FIPS_CONFIG);
-			fips |= ICA_FIPS_INTEGRITY;
+			fips = ICA_FIPS_INTEGRITY;
```
jschmidb, line 297: Why did you change the |= into = ? The fips variable could contain the ICA_FIPS_MODE flag from the new fips_get_indicator function. So we would overwrite this here. However, in case of ICA_FIPS_INTEGRITY the library would stop working anyway.

ifranzki: The old code did set fips = 0 in line 284.  Then in case of an error it or'ed ICA_FIPS_INTEGRITY onto it. The result is fips = ICA_FIPS_INTEGRITY in error case.
At the end of the if block (good case), it or'ed ICA_FIPS_MODE onto it.
So in good case fips = ICA_FIPS_MODE and in bad case fips = ICA_FIPS_INTEGRITY .

That whats the new code does, too, to, just less complicated :-) 

On entry of the block fips =ICA_FIPS_MODE, and in case of error I set it to fips = ICA_FIPS_INTEGRITY.
If no error occurred, it remains at fips =ICA_FIPS_MODE,

jschmidb (approved):
lgtm

PR #124, opened 2024-01-19
Various fixes including fixes for protecting buffers containing sensitive data

jschmidb:


src/ica_api.c
```diff
@@ -1200,27 +1200,29 @@ unsigned int ica_rsa_crt_key_check(ica_rsa_key_crt_t *rsa_key)
...
-		/* calculate new qInv */
-		bn_p = BN_new();
-		bn_q = BN_new();
-		bn_invq = BN_new();
+		bn_p = BN_secure_new();
+		bn_q = BN_secure_new();
+		bn_invq = BN_secure_new();
 		ctx = BN_CTX_new();
+		if (!bn_p || !bn_q || !bn_invq || !ctx)
+			return ENOMEM;
```
ifranzki, line 1212: You need to free the BIGNUMs and the CTX in this error case.
At least one if it is NULL, but the others might be non-NULL and thus must be freed.

src/ica_api.c
```diff
@@ -1237,6 +1239,7 @@ unsigned int ica_rsa_crt_key_check(ica_rsa_key_crt_t *rsa_key)
```
ifranzki, line 1239: OPENSSL_cleanse for tmp_buf before it is freed in line 1231? It dies contain private key parts.

src/ica_api.c
```diff
@@ -1273,7 +1276,9 @@ unsigned int ica_rsa_crt(ica_adapter_handle_t adapter_handle,
 	rb.outputdata = output_data;
 	rb.outputdatalength = rsa_key->key_length;
 
-	ica_rsa_crt_key_check(rsa_key);
+	rc = ica_rsa_crt_key_check(rsa_key);
+	if (rc > 1)
+		return rc;
```
ifranzki, line 1281: in line 1310 you cleans rb before returning. Should we also cleanse it here in that error case ?

src/s390_ecc.c
```diff
@@ -1268,7 +1269,10 @@ static unsigned int provide_pubkey(const ICA_EC_KEY *privkey, unsigned char *X,
 	}
 
 	/* Get (D) as BIGNUM */
-	if ((bn_d = BN_bin2bn(privkey->D, privlen, NULL)) == NULL) {
+	bn_d = BN_secure_new();
+	if (bn_d == NULL)
+		return ENOMEM;
+	if (BN_bin2bn(privkey->D, privlen, bn_d) == NULL) {
 		return EFAULT;
```
ifranzki, line 1276: free bn_d here in this error case.

src/ica_api.c
```diff
@@ -1200,27 +1200,29 @@ unsigned int ica_rsa_crt_key_check(ica_rsa_key_crt_t *rsa_key)
...
+		memcpy(tmp_buf, rsa_key->p + 8, rsa_key->key_length / 2);
+		memcpy(rsa_key->p + 8, rsa_key->q, rsa_key->key_length / 2);
+		memcpy(rsa_key->q, tmp_buf, rsa_key->key_length / 2);
+
+		/* swap dp and dq */
+		memcpy(tmp_buf, rsa_key->dp + 8, rsa_key->key_length / 2);
+		memcpy(rsa_key->dp + 8, rsa_key->dq, rsa_key->key_length / 2);
+		memcpy(rsa_key->dq, tmp_buf, rsa_key->key_length / 2);
+
+		BN_bin2bn(rsa_key->p, rsa_key->key_length / 2 + 8, bn_p);
```
ifranzki, line 1224: While you are at it: Error checking for the BN_bin2bn calls?

ifranzki: as well as for BN_mod_inverse and BN_bn2binpad further down

src/ica_api.c
jschmidb: The sensitive information is copied into rb below this return, so I think we can leave it as it is.

src/ica_api.c
jschmidb: Yes, right. I changed the logic a bit so that we always run through the free / clear_free.

src/ica_api.c
jschmidb: Done.

src/s390_ecc.c
jschmidb: Done.

src/ica_api.c
jschmidb: Done.

src/ica_api.c
ifranzki: Actually `rb` only contains pointers, so it does not contain any sensitive information anyway.

ifranzki (approved):

Issue #126, opened 2024-04-15
adapter_handle_test fails when CEX adapter is available


sharkcz:
If I read the test right, then the `adapter_handle_test` expects that opening `/dev/z90crypt` device will return `3` as the file descriptor. But it doesn't seem to be the case when a CEX adapter is available in the system as something does some kind of initialization and grabs the fd `3`. The output is following
````
[root@s390x-kvm-121 libica]# LD_LIBRARY_PATH=./src/.libs ./test/adapter_handle_test -v
ica_open_adapter: file descriptor value is greater than 3 (current 4).
````
`strace` than reveals
````
[root@s390x-kvm-121 libica]# LD_LIBRARY_PATH=./src/.libs strace ./test/adapter_handle_test -v
execve("./test/adapter_handle_test", ["./test/adapter_handle_test", "-v"], 0x3ffdb4f9308 /* 44 vars */) = 0
brk(NULL)                               = 0x1cd5000
mmap(NULL, 8192, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x3ff84ef9000
...
mmap(NULL, 8192, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x3ff84ef7000
set_tid_address(0x3ff84ef8350)          = 58839
set_robust_list(0x3ff84ef8360, 24)      = 0
rseq(0x3ff84ef89a0, 0x20, 0, 0xb2ff0fff) = 0
mprotect(0x3ff848e3000, 12288, PROT_READ) = 0
mprotect(0x3ff8469b000, 4096, PROT_READ) = 0
mprotect(0x3ff84caa000, 348160, PROT_READ) = 0
mprotect(0x3ff84e33000, 4096, PROT_READ) = 0
mprotect(0x1003000, 4096, PROT_READ)    = 0
mprotect(0x3ff84eac000, 8192, PROT_READ) = 0
prlimit64(0, RLIMIT_STACK, NULL, {rlim_cur=8192*1024, rlim_max=RLIM64_INFINITY}) = 0
munmap(0x3ff84d80000, 17579)            = 0
openat(AT_FDCWD, "/dev/z90crypt", O_RDWR|O_CLOEXEC) = 3
ioctl(3, ICARSAMODEXPO, 0x3ffd64f9730)  = 0
[...]

PR #127, opened 2024-04-16
test: disable CEX usage in OpenSSL for all tests

holger-dengler:
OpenSSL supports CEX exploitation since version v3.2.x. Libica and its testcases use OpenSSL as helper and fallback, so disable the CEX acceleration for all tests.

If the environment variable is already set, use it as is without modifying it. In this case, it is up to the user to choose the right settings.

Fixes: Issue #126
Link: https://github.com/opencryptoki/libica/issues/126

### DIFF
--- a/test/Makefile.am
+++ b/test/Makefile.am
@@ -61,10 +61,14 @@ TESTS += \
 ${top_builddir}/src/internal_tests/ec_internal_test
 endif
 
+# disable OpenSSL CEX usage for all tests
+OPENSSL_s390xcap ?= nocex
+
 TEST_EXTENSIONS = .sh .pl
 TESTS_ENVIRONMENT = export LD_LIBRARY_PATH=${builddir}/../src/.libs/:$$LD_LIBRARY_PATH \
 			   PATH=${builddir}/../src/:$$PATH \
-			   LIBICA_TESTDATA=${srcdir}/testdata/;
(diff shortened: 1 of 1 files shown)


ifranzki (dismissed):

sharkcz:
Running `make check` still fails here (current master + this PR). Running the test manually with `OPENSSL_s390xcap=nocex LD_LIBRARY_PATH=./src/.libs ./test/adapter_handle_test -v` makes it pass, so technically setting the `OPENSSL_s390xcap` variable is the right approach.

sharkcz:
hmm, should `OPENSSL_s390xcap` be set to `noexec` or to `nocex`, seems the test fails with `noexec`

test/Makefile.am
```diff
@@ -61,10 +61,14 @@ TESTS += \
 ${top_builddir}/src/internal_tests/ec_internal_test
 endif
 
+# disable OpenSSL CEX usage for all tests
+OPENSSL_s390xcap ?= noexec
```
ifranzki, line 65: should be `nocex`

holger-dengler: uups, sorry

holger-dengler: change has been force-pushed

sharkcz:
and it looks good now, thanks

sharkcz (approved):
LGTM

ifranzki (approved):

PR #129, opened 2024-06-19
Use __asm__  instead of asm

jschmidb:
The asm keyword is a GNU extension. When writing code that can be compiled with -ansi and the various -std options, use __asm__ instead of asm.


ifranzki (approved):
I haven't checked if you found all of them, but I guess you did a global search & replace...

holger-dengler (approved):
Looks good to me.

PR #132, opened 2024-08-27
TRAVIS: Also build it with clang

ifranzki:


ifranzki:
@jschmidb Did you intentionally not include this into 4.3.1 ? I think the "Fix s390_kimd/kmld_shake inline assemblies" commit is worth to have in any case....

jschmidb:
> Did you intentionally not include this into 4.3.1 ? 

Yes, because I did not test yet. But it will be included with the next couple of fixes.

PR #138, opened 2024-12-12
Fips addition

jschmidb:
Introduce a new function list property flag to indicate that no small (< 64k) RSA public exponents are allowed.
Also show the minimum allowed public exponent in the icainfo output


ifranzki (approved):
lgtm

PR #140, opened 2025-04-08
CCA adapters are not usable in an Secure Execution guest

ifranzki:
Trying to use CCA adapters in an Secure Execution guest fails, because CCA adapters are not usable there. Nevertheless they show up as online, so special checking needs to be added to detect this situation.

Don't set the 'ecc_via_online_card' flag in such case, so that the EC mechanisms are reported to not be available with dynamic hardware when running in an Secure Execution guest, and attempts to use curves that would require a CCA adapter fail with ENODEV. This can then be handled by the calling applications appropriately.

### DIFF
--- a/src/s390_crypto.c
+++ b/src/s390_crypto.c
@@ -403,7 +403,11 @@ unsigned int search_for_cards()
 	char buf[250];
 	struct dirent *direntp;
 	char type[6];
-	int rc;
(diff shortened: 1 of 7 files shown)


jschmidb (approved):
lgtm, thanks for fixing!

PR #141, opened 2025-04-28
Fix declaration of ica_aes_gcm_kma_ctx_new in ica_api.h header

ifranzki:
When building an application that includes ica_api.h with -Wstrict-prototypes, then this shows warnings/errors for the ica_aes_gcm_kma_ctx_new() declaration in ica_api.h.

The declaration in ica_api.h is anyway different that the one used in ica_api.c:

ica_api.h:
```
  ICA_EXPORT
  kma_ctx* ica_aes_gcm_kma_ctx_new();
```

ica_api.c:
`  kma_ctx* ica_aes_gcm_kma_ctx_new(void)`

### DIFF
--- a/include/ica_api.h
+++ b/include/ica_api.h
@@ -3370,7 +3370,7 @@ typedef struct kma_ctx_t kma_ctx;
  * NULL if no memory could be allocated.
  */
 ICA_EXPORT
-kma_ctx* ica_aes_gcm_kma_ctx_new();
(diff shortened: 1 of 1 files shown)


ifranzki:
It was the only place in the external header file. 

Inside libica code there are more places with empty arguments functions. As long as libica itself is not built with  `-Wstrict-prototypes` this does not hurt.

ifranzki:
@jschmidb do you want to merge this?

PR #143, opened 2025-06-04
CONFIGURE: Make the OpenSSL FIPS config file name configurable

ifranzki:
The name of the OpenSSL FIPS config file may be different on various distros. It is included in src/openssl3-fips.cnf when used with OpenSSL 3.0 or later.

To use a specific name:
  ./configure --enable-fips --with-fips-config=fips_local.cnf

The default remains fipsmodule.cnf. It is only used when --enable-fips is also specified, and libica is built against OpenSSL 3.0 or later.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -100,6 +100,13 @@ if test "x$enable_fips" = xyes; then
     fi
 fi
 
+dnl --- with-fips-config
+AC_ARG_WITH([fips-config],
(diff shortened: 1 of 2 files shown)


jschmidb (approved):
lgtm

ifranzki:
Feel free to merge whenever you are ready :-)

PR #145, opened 2026-05-05
Block SHA1 mechanism for FIPS 140-3:

vkanjur:
This commit removes SHA1 mechanism when FIPS 140-3 is active.

### DIFF
--- a/src/fips.c
+++ b/src/fips.c
@@ -156,7 +156,6 @@ static int sha##_sha_##_kat(void) {					\
 	}								\
 	return 0;							\
 }
-SHA_KAT(1, );
 SHA_KAT(224, 256);
 SHA_KAT(256, 256);
 SHA_KAT(384, 512);
@@ -180,7 +179,6 @@ static int sha##_sha_##_kat(void) {					\
 	}								\
 	return 0;							\
 }
-SHA_KAT(1, );
 SHA_KAT(224, 256);
 SHA_KAT(256, 256);
 SHA_KAT(384, 512);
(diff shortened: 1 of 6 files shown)


src/fips.c
```diff
@@ -1373,7 +1373,7 @@ int FIPS_BLACKLIST[] = {DES_ECB, DES_CBC, DES_CBC_CS, DES_OFB,
 	DES3_CBC, DES3_CBC_CS, DES3_OFB, DES3_CFB, DES3_CTR, DES3_CTRLST,
 	DES3_CBC_MAC, DES3_CMAC, ED25519_KEYGEN, ED25519_SIGN, ED25519_VERIFY,
 	ED448_KEYGEN, ED448_SIGN, ED448_VERIFY, X25519_KEYGEN, X25519_DERIVE,
-	X448_KEYGEN, X448_DERIVE, RSA_ME, RSA_CRT, SHA512_DRNG, -1, -1 };
+	X448_KEYGEN, X448_DERIVE, RSA_ME, RSA_CRT, SHA512_DRNG, SHA1, -1, -1 };
```
ifranzki, line 1374: Adding SHA1 here causes it to no longer be included in the mechanism list (`ica_get_functionlist()`), but it does not hinder anyone from using the `ica_sha1()` function.....

I checked how this is done for e.g. DES, and there the ica_des_xxx() functions have
```
#ifdef ICA_FIPS
	if (fips)
		return EACCES;
#endif /* ICA_FIPS */
```
at the top of the function body. So this block the function when fips is active. 

Note that ica_sha1() has the following there:

```
#ifdef ICA_FIPS
	if (fips >> 1)
		return EACCES;
#endif /* ICA_FIPS */
```

This blocks the function only if the fips selftests have failed (note `(fips)` vs `(fips >> 1)`).

ifranzki (approved):
lgtm

PR #146, opened 2026-06-12
Fips changes

vkanjur:


include/ica_api.h
```diff
@@ -83,7 +83,7 @@ typedef ica_adapter_handle_t ICA_ADAPTER_HANDLE;
 #define ICA_PROPERTY_AES_256		0x00000004
 
 #define ICA_PROPERTY_RSA_ALL		0x0000000F /* All RSA key lengths */
-#define ICA_PROPERTY_RSA_FIPS		0x0000000C /* RSA 2k and higher */
+#define ICA_PROPERTY_RSA_FIPS		0x00020000 /* RSA 3K and higher */
```
ifranzki: This one is not correct.

The 3K bit must also be added to ICA_PROPERTY_RSA_ALL (for non-FIPS).
For ICA_PROPERTY_RSA_FIPS, the bits for 3K and 4K must be set, not only the 3k bit.

I would suggest to do it as follows, this makes it better readable:

```
#define ICA_PROPERTY_RSA_512		0x00000001
#define ICA_PROPERTY_RSA_1024		0x00000002
#define ICA_PROPERTY_RSA_2048		0x00000004
#define ICA_PROPERTY_RSA_4096		0x00000008
#define ICA_PROPERTY_RSA_3072		0x00020000
#define ICA_PROPERTY_RSA_ALL		(ICA_PROPERTY_RSA_512  | \
					 ICA_PROPERTY_RSA_1024 | \
					 ICA_PROPERTY_RSA_2048 | \
					 ICA_PROPERTY_RSA_3072 | \
					 ICA_PROPERTY_RSA_4096)
#define ICA_PROPERTY_RSA_FIPS		(ICA_PROPERTY_RSA_3072 | \
					 ICA_PROPERTY_RSA_4096)
```

You might decide to define ICA_PROPERTY_RSA_3072 as 0x00000010 instead of 0x00020000, i.e. to just extend the existing size bits on the left. However, it might be a bit strange that the 4K bit is before the 3K bit (we can't change the existing bits for compatbility reasons). Also, if we ever add bits for higher sizes (e.g. 8K), we would then add the 8K bit on the left of the 3K bit ....

Anyway. its a mess either way, so feel free to decide whatever you want :-)

vkanjur: It makes it readable. I will fix it.

src/fips.c
```diff
@@ -1371,16 +1376,16 @@ int FIPS_BLACKLIST[] = {DES_ECB, DES_CBC, DES_CBC_CS, DES_OFB,
...
+	SHA512_DRNG, SHA1, AES_ECB, SHA224, SHA512_224, SHA3_224, -1, -1 };
 const size_t FIPS_BLACKLIST_LEN
 	= sizeof(FIPS_BLACKLIST) / sizeof(FIPS_BLACKLIST[0]);
 
 /*
  * FIPS service indicator: List of tolerated but non-approved algorithms.
  */
 int FIPS_OVERRIDE_LIST[] = { RSA_ME, RSA_CRT, SHA512_DRNG, AES_ECB, SHA224,
-	SHA512_224, SHA3_224, -1, -1 };
+	SHA512_224, SHA3_224, RSA_KEY_GEN_ME, RSA_KEY_GEN_CRT, -1, -1 };
```
ifranzki, line 1391: Just a nit, feel free to ignore :-) 

For FIPS_BLACKLIST you added the RSA key gens right after RSA_ME and RSA_CRT, but here you added the key gens after the digests.... Any particular reason for this?

ifranzki (approved):
LGTM

ifranzki:
Travis seems to have problems booting the s390x systems. I think that's just the 'normal' Travis hiccup... Not related to the code changes.

PR #148, opened 2026-07-07
fips_powerup_tests: extend RSA KAT coverage to 3072 and 4096 bits

vkanjur:
Add RSA known-answer test vectors for 3072-bit and 4096-bit key sizes to improve rsa_kat coverage during FIPS power-up self-tests.


ifranzki (approved):
LGTM.

I assume you did run it with FIPS enabled to see if the new tests really run. In Travis it builds with FIPS enabled, but the tests are without the system being in FIPS mode, so it won't run the FIPS self tests.

vkanjur:
yes I tested with FIPS enabled.

PR #150, opened 2026-07-31
Minor fixes

vkanjur:


ifranzki (approved):
LGTM.

The only thing is that `s390_kmo()` which is called from `s390_des_ofb_hw()` and `s390_aes_ofb_hw()` uses a signed long as `src_len`. However, it just passes this as-is to the register and to the instruction, so it should not matter much for the instruction execution. Only for the return statement it does some calculations with `src_len` which could cause over-/underfllows when `src_len` is large enough so that the high order bit is on. 

So we might want to change s390_kmo's `src_len` parameter to unsigned long as well. However, this might be the same with the other inline assemblies.... It might also be needed to change the return type of these functions then. I guess we should carefully think this through and maybe do it in a separate PR.

PR #151, opened 2026-08-07
icastats: Do not overflow statistics counters, saturate at max

ifranzki:
When statistic counters would overflow, keep them at the maximum possible value. Wrapping to zero is the worst possible outcome for an auditing tool, it silently under-reports activity.

If a counter saturated, then the icastats display indicates this with "[saturated]" instead of the counter value. JSON output will have an additional boolean field with the suffix '-saturated' set to 'true' emitted alongside the counter value.

Also make the reset operation to operate in an per counter atomic way. Otherwise counter increments might get lost while a reset is performed.

While at it, escape special charterers in JSON strings, as required by RFC 8259.

### DIFF
--- a/doc/icastats.1
+++ b/doc/icastats.1
@@ -68,6 +68,16 @@ or system shut down. This also means that the statistical data shown with
 icastats is on a per user base and only the root user is able to see and
 maintain the collection of statistic data from libica at system scope.
 .P
+Counter values saturate at their maximum (UINT64_MAX = 18446744073709551615)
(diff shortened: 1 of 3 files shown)


vkanjur (approved):
lgtm

ifranzki:
wait a bit with merging, there might be additional changes needed....

ifranzki:
Some changes for the summary calculation were missing. Should be fixed now.

vkanjur:
lgtm..merging!

